PR #4, opened 2018-11-16
Parse verification unit binding indications

FenrisTC:
This change:
 * adds the ability to parse verification unit binding indications for configuration declarations (LRM 3.4), component configurations (LRM 3.4.3), and compound configuration specifications (LRM 7.3).
 * adds the 'vunit' keyword as a token kind.
 * makes the 'parse_use_clause_no_keyword' function public inside the crate. I did this to be able to parse the use_clause in parse_configuration_declation less awkwarldy.

### DIFF
--- a/vhdl_parser/src/ast.rs
+++ b/vhdl_parser/src/ast.rs
@@ -928,12 +928,34 @@ pub struct ComponentSpecification {
     pub component_name: SelectedName,
 }
 
-/// LRM 7.3 Configuration specification
-// @TODO There is also compound
+
+/// LRM 7.3.4 Verification unit binding indication
 #[derive(PartialEq, Debug, Clone)]
-pub struct ConfigurationSpecification {
+pub struct VUnitBindingIndication {
+    pub vunit_list: Vec<WithPos<Name>>,
+}
+
+
+/// LRM 7.3 Compound Configuration specification
+#[derive(PartialEq, Debug, Clone)]
+pub struct CompoundConfigurationSpecification {
     pub spec: ComponentSpecification,
     pub bind_ind: BindingIndication,
+    pub vunit_bind_inds: Vec<VUnitBindingIndication>,
+}
+
+/// LRM 7.3 Simple Configuration specification
+#[derive(PartialEq, Debug, Clone)]
+pub struct SimpleConfigurationSpecification {
+    pub spec: ComponentSpecification,
+    pub bind_ind: BindingIndication,
+}
+
+/// LRM 7.3 Configuration specification
+#[derive(PartialEq, Debug, Clone)]
+pub enum ConfigurationSpecification {
+    Simple(SimpleConfigurationSpecification),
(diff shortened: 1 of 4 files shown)


vhdl_parser/src/ast.rs
```diff
@@ -928,12 +928,34 @@ pub struct ComponentSpecification {
...
     pub bind_ind: BindingIndication,
+    pub vunit_bind_inds: Vec<VUnitBindingIndication>,
+}
+
+/// LRM 7.3 Simple Configuration specification
+#[derive(PartialEq, Debug, Clone)]
+pub struct SimpleConfigurationSpecification {
+    pub spec: ComponentSpecification,
+    pub bind_ind: BindingIndication,
+}
```
kraigher: Maybe it is unnecessary with two struct here. A simple configuration could just be a compound with and empty vunit_bind_indication. In this way the data model becomes a bit simpler.

vhdl_parser/src/ast.rs
```diff
@@ -948,6 +970,7 @@ pub enum ConfigurationDeclarativeItem {
 pub struct ComponentConfiguration {
     pub spec: ComponentSpecification,
     pub bind_ind: Option<BindingIndication>,
+    pub vunit_bind_inds: Option<Vec<VUnitBindingIndication>>,
```
kraigher: Maybe Option<Vec> could just be a Vec. None would be an empty Vec.

vhdl_parser/src/ast.rs
```diff
@@ -972,6 +995,7 @@ pub struct ConfigurationDeclaration {
     pub ident: Ident,
     pub entity_name: SelectedName,
     pub decl: Vec<ConfigurationDeclarativeItem>,
+    pub vunit_bind_inds: Option<Vec<VUnitBindingIndication>>,
```
kraigher: Maybe Option<Vec> could just be a Vec. None would be an empty Vec.

kraigher:
Looks good I just had a few comments where I think the ast could be simpler. I will look into why the CI fails.

kraigher:
Ok the build fails because the code is not `cargo fmt` formatted.

vhdl_parser/src/ast.rs
FenrisTC: Right, the two structs seem similar enough. I'll pack them together.

vhdl_parser/src/ast.rs
FenrisTC: Ok, makes sense. I'll remove it.

vhdl_parser/src/ast.rs
FenrisTC: Ok, makes sense. I'll remove it.

FenrisTC:
> Ok the build fails because the code is not `cargo fmt` formatted.

Riiight. I somewhat forgot cargo fmt existed.

PR #12, opened 2018-11-29
Add shutdown method and exit notification to language server

mbrobbel:
This adds the shutdown method and exit notification to the language server to prevent errors when using this with a client which use this method and notification to stop the server.
Can you please review? I'm not sure if this correctly stops the spawned threads and closes the channels.

### DIFF
--- a/vhdl_ls/src/stdio_server.rs
+++ b/vhdl_ls/src/stdio_server.rs
@@ -80,7 +80,10 @@ pub fn start() {
             let request = read_request(&mut stdin.lock());
             match request_sender.send(request) {
                 Ok(_) => continue,
-                Err(_) => break,
+                Err(_) => {
+                    eprintln!("Channel hung up. Unlocking stdin handle.");
+                    break;
+                }
             }
         }
     });
@@ -94,6 +97,7 @@ pub fn start() {
                     send_response(&mut stdout, &response);
                 }
                 Err(_) => {
+                    eprintln!("Channel hung up.");
                     break;
                 }
             }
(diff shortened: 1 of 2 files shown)


vhdl_ls/src/stdio_server.rs
```diff
@@ -64,24 +78,39 @@ pub fn start() {
         let stdin = io::stdin();
         loop {
             let request = read_request(&mut stdin.lock());
-            request_sender.send(request).unwrap();
+            match request_sender.send(request) {
+                Ok(_) => continue,
+                Err(_) => break,
```
kraigher: Maybe add an eprintln! of the error to not silently swallow it

vhdl_ls/src/stdio_server.rs
```diff
@@ -64,24 +78,39 @@ pub fn start() {
...
     spawn(move || {
         let mut stdout = io::stdout();
         loop {
-            let response: String = response_receiver.recv().unwrap();
-            send_response(&mut stdout, &response);
+            match response_receiver.recv() {
+                Ok(response) => {
+                    send_response(&mut stdout, &response);
+                }
+                Err(_) => {
```
kraigher: Maybe add an eprintln! of the error to not silently swallow it

vhdl_ls/src/stdio_server.rs
```diff
@@ -64,24 +78,39 @@ pub fn start() {
...
-            response_sender.send(response).unwrap();
+        match request_receiver.recv() {
+            Ok(request) => {
+                let response = io.handle_request_sync(&request);
+                if let Some(response) = response {
+                    response_sender.send(response).unwrap();
+                }
+            }
+            Err(_) => {
+                break;
```
kraigher: Maybe add an eprintln! of the error to not silently swallow it

PR #15, opened 2018-12-02
Fix some clippy warnings (part 2)

mbrobbel:


mbrobbel:
Please don't merge yet. There are some issues.

vhdl_ls/src/stdio_server.rs
```diff
@@ -134,10 +134,10 @@ fn read_request(reader: &mut BufRead) -> String {
 
 fn send_response(writer: &mut Write, response: &str) {
     trace!("SEND RESPONSE: {:?}", response);
-    write!(writer, "Content-Length: {}\r\n", response.len());
-    write!(writer, "\r\n");
+    writeln!(writer, "Content-Length: {}", response.len());
```
kraigher: The protocol requires it to be \r\n, writeln will just insert \n on most platforms.

vhdl_parser/src/tokenstream.rs
```diff
@@ -76,7 +76,7 @@ impl TokenStream {
         Ok(self.peek()?.map(|ref token| token.kind))
     }
 
-    pub fn is_peek_kinds(self: &mut Self, kinds: &[Kind]) -> ParseResult<bool> {
+    pub fn get_peek_kinds(self: &mut Self, kinds: &[Kind]) -> ParseResult<bool> {
```
kraigher: I do not think `get_` is a good prefix for this function. Maybe rename it to `next_kinds_are`. It is just a check if the next sequence of tokens in the stream match the sequence of kinds in the argument. It is not *getting* anything.

mbrobbel: 054e4ec

PR #23, opened 2018-12-09
WIP: Parse comments

benreynwar:
This is nowhere near being ready for merging but I'm creating a merge request so that I can get some feedback.

Currently it's just handling the tokenizing of the comments, adding them as meta data into the other tokens.  Comments at the end of the file are not yet handled.

For handling the final comments I'm thinking of either:
1) Adding a special token to hold the final comments or,
2) Add an additional method to the tokenizer to retrieve the final comments.

Once that's done the next step is moving the comments from the tokens into the AST.

### DIFF
--- a/vhdl_parser/src/tokenizer.rs
+++ b/vhdl_parser/src/tokenizer.rs
@@ -6,7 +6,7 @@
 
 use self::fnv::FnvHashMap;
 use crate::message::{Message, ParseResult};
-use crate::source::{Source, SrcPos, WithPos};
+use crate::source::{Pos, Source, SrcPos, WithPos};
 use fnv;
 
 use crate::ast;
@@ -401,6 +401,19 @@ pub struct Token {
     pub kind: Kind,
     pub value: Value,
     pub pos: SrcPos,
+    pub comments: Option<Box<TokenComments>>,
+}
+
+#[derive(PartialEq, Clone, Debug)]
+pub struct TokenComments {
+    pub leading: Vec<Comment>,
+    pub trailing: Option<Comment>,
+}
+
+#[derive(PartialEq, Clone, Debug)]
+pub struct Comment {
+    pub value: Latin1String,
+    pub pos: Pos,
 }
 
 use std::convert::AsRef;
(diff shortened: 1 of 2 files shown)


vhdl_parser/src/tokenizer.rs
```diff
@@ -401,6 +401,14 @@ pub struct Token {
     pub kind: Kind,
     pub value: Value,
     pub pos: SrcPos,
+    pub leading_comments: Vec<Comment>,
+    pub trailing_comment: Option<Comment>,
```
kraigher: Maybe the comments should be put in a struct and put in an Option<Box<Commens>> to minimize Token sizes when there are no comments.

vhdl_parser/src/tokenizer.rs
```diff
@@ -401,6 +401,14 @@ pub struct Token {
...
+}
+
+#[derive(PartialEq, Clone, Debug)]
+pub struct Comment {
+    pub value: Latin1String,
+    pub pos: SrcPos,
```
kraigher: The start, length fields of the SrcPos should be factored out to their own struct Position and this struct can be used here. Then we can avoid cloning the Arc.

vhdl_parser/src/tokenizer.rs
kraigher: Also consider move_after(&token) when there are trailing comments.

benreynwar: I put the token comments in an Option<Box> as suggested.

I've modified move_after so I can use it with trailing_comments,  but I haven't pushed that yet.  I get an error about mutable and immutable borrows, which I need to work out.  I'm enjoying rust, but haven't quite got the hang of it yet!

vhdl_parser/src/tokenizer.rs
benreynwar: I made this change, but I don't understand why it was necessary.  Every Token already has a reference to source in its 'pos' field.  Why is it different if a Comment also holds a reference to source?

kraigher: Maybe it is a micro optimization but it is an extra clone. Cloning an Arc requires some kind of thread synchronization so it is something to watch out for. Maybe you can leave it as it is. I am thinking of storing a reference to the source instead in the future to avoid all clones.

vhdl_parser/src/tokenizer.rs
```diff
@@ -401,14 +401,27 @@ pub struct Token {
...
 #[derive(PartialEq, Clone, Debug)]
 pub struct Comment {
     pub value: Latin1String,
-    pub pos: SrcPos,
+    pub pos: Pos,
+}
+
+// Like SrcPos but without Source so we don't need to clone it
+// all the time.
+#[derive(PartialEq, Clone, Debug)]
```
kraigher: This would belong in source.rs and be used by SrcPos.

PR #24, opened 2018-12-10
Error recovery

FenrisTC:
This is a first step to recover from parse errors.
It adds recovery from declarative parts, as well as some infrastructure to make recovery somewhat ergonomic.


kraigher:
Could you split this pull request into two parts, one with the sensitivity list and block configuration and another with the recovery. I am ok with the first part but the recovery part I have some issues with.

kraigher:
Regarding recovery in declarative parts I will open an issue and we can have a discussion on the requirements and challenges before we commit to implementing a solution.

kraigher:
I made an issue for recovery in declarative parts: https://github.com/kraigher/rust_hdl/issues/25

FenrisTC:
@kraigher I opened a second pull request only containing the bug fixes.
I'll keep this one open as a more general WIP / error recovery pull request to track progress.

vhdl_parser/src/declarative_part.rs
```diff
@@ -82,46 +83,69 @@ pub fn parse_declarative_part_leave_end_token(
...
-                for decl in parse_object_declaration(stream)? {
-                    declarations.push(Declaration::Object(decl))
-                }
-            }
-            File => {
-                for decl in parse_file_declaration(stream)? {
-                    declarations.push(Declaration::File(decl))
+            Begin | End => break,
+
+            kind if multi_decl_tokens.contains(&kind) => {
```
kraigher: Using Vec.contains is O(n) which is not ideal. It is better to create an `is_<xyz>_kinds` function that uses a match statement internally to return true/false

vhdl_parser/src/declarative_part.rs
```diff
@@ -176,6 +199,25 @@ package ident is new lib.foo.bar
...
+",
+        );
+        let (decls, msgs) =
+            code.with_partial_stream_messages(parse_declarative_part_leave_end_token);
+        let decls = decls.unwrap();
+        assert_eq!(decls.len(), 1);
+        match decls[0] {
+            Declaration::Object(..) => (),
+            _ => panic!("expected Object declaration, got {:?}", decls[0]),
+        }
```
kraigher: I would prefer if the decls and error messages where tested in full. The `code` utility can be used to create the expected values conveniently.

vhdl_parser/src/tokenstream.rs
```diff
@@ -102,6 +102,17 @@ impl TokenStream {
         Ok(self.pop_if_kind(kind)?.is_some())
     }
 
+    pub fn skip_to(self: &mut Self, kinds: &[Kind]) -> ParseResult<()> {
```
kraigher: better make this an `Fn` that returns true or false depending on if the kind is correct or not. Vec::contains is O(n)

FenrisTC: I guess I could adapt it to have a closure as a second parameter, to feed the match statement from each call site into it without duplicating it for each distinct list of recovery tokens.

From a performance point of view, it's not clear to me that a O(n) linear search is actually slower than jump tables.
We would need multiple jump tables, each one with >100 entries to match on all possible values of the token kind.
The list to search on the other hand would be at most around 20 elements large and I think the typical case will be around 1-2 Elements (e.g. recovering to a semicolon).

kraigher: I think it gets implementet as a shift and bitwise and for enums less than 128 kinds.

vhdl_parser/src/tokenstream.rs
```diff
@@ -127,6 +137,46 @@ impl TokenStream {
...
+    ) -> ParseResult<T> {
+        if self.is_ok() {
+            return self;
+        }
+
+        let res = stream.skip_until(cond);
+        match res {
+            Ok(_) => self,
+            Err(err) => {
+                msgs.push(self.unwrap_err());
```
kraigher: Isnt the message pushed twice? First here and then after the call to or_recover_until above?

kraigher: I think there is a test missing here. The test if implemented would show that the message is pushed twice.

kraigher:
Is it ready to merge or are you planning to include more stuff in the PR (The title says WIP)? I am ready to approve it.

FenrisTC:
It's ready to merge, sorry.

PR #29, opened 2018-12-28
WIP: Add listener mod for traversing the AST

Bochlin:
Concept for an AST listener, making a pull request to discuss whether this is a suitable addition.

The basic concept is that is should be possible to implement a listener for the AST for a specific function (e.g. a code generator or linter). The listener is borrowed by a notifier which is then passed to a listen function defined in a trait. While the notifier traverses the AST it calls enter and exit methods (events) in the listener defined for each, it is possible to register multiple listeners with the notifier so that the AST only needs to be traversed once.

The concept is currently implemented for a small set of nodes.

@kraigher Do you think this kind of approach fits your model?


kraigher:
The ultimate criteria if it fits or not is if it is useful in at least two scenarios. For features that only want to look at some parts of the AST without maintaining logic for traversing the entire AST this design pattern makes sense. 

For features that will want to look at the entire AST anyway like semantic analysis it is probably easier to write custom traversal functions where the state can be maintained in local variables rather than having to mutate the self to record enter and exit of AST nodes.

Bochlin:
I agree that semantic analysis probably is easier to do with custom traversal for most parts. My thoughts with the listener is that it can run its pass over an AST where the model is "linked", meaning that implementers of a listener can follow references around the model if they'd like. This would require that some semantic analysis has been done already. Some other checks, like checking that assignments are not made to input ports would suit this design pattern in my opinion as we want to do the same check for each assignment, regardless of where it is in the hierarchy. 

The use cases below are to me cases where a listener would be useful for traversing the entire AST while being more or less state-less except for the stored messages. They do however depend on being able to either access the symbol table directly, or even better, have a fully linked AST where the links can be followed.

**Writing an XML representation**
When writing the XML, we want to write according to some pattern for each design element, Example use of events:
enter_(entity_declaration/architecture or other design unit): Open file and name it based on design unit name and kind, write header and open tag for design_unit xml element.
enter_(some declaration): Write declaration to xml, including both open and close tags for the declaration.
exit_design unit: Write close tag and close the file.

**LINT**
The checks for each design element is implemented in an event and violations are written to a Vec<Message> in the listener struct. To me it is easier to use a listener here since it would remove any need for knowledge about the actual hierarchy to find for example all signal declarations within a design unit (doesn't matter if they are in a generate, block or the declarative part of an entity or architecture). Implementing a checker for e.g. signal declarations would be done by adding logic to a enter_signal_declaration event.

**Converting VHDL-2008 constructs to VHDL-93 compatible constructs**
You could implement a listener that converts all sequential conditional signal assignments to if-then statements by providing an implementation for a enter_conditional_signal_assignment event. Main benefit again would be that the implementation does not need to go look for the assignments as this is  provided by the notifier in the form of events.

kraigher:
Why is there no mutable reference to the AST in the exit methods?

kraigher:
I need something similar now to search for an ast object given a source file position.
I think this use case requires something different like a Searcher trait because you do not want to recurse down into the tree when you found a match. Maybe a Listener can be implemented on top of a Searcher configured to always recurse deeper into the tree.

Bochlin:
I had too take a look and see how far I got with this. I had renamed the listener to visitor and implemented it for almost the full AST. I also started experimenting with the listener keeping track of the source position but I did not complete that. I have pushed what I did to my fork now if you want to see the current state. The way I did it allowed for multiple visitors to be added to a walk of the AST, but I would like to simplify that so that only a single listener is used.

Wouldn't a searcher be an implementation of the listener/walker? It should be possible to add functionality to the walker to abort, that way the searcher could abort when it has found a match. I'm not sure however how to find an entry point for the visitor given a source position.

kraigher:
It is semantics; I you generalize the Visitor it can become what I meant by a Searcher. If we do not want a generalized Visitor a Searcher can be created upon which a less general Visitor is implemented. Maybe it is better to make the Visitor more general.
 
The visitor does not need to know the entry point for the source position. Its job would be to find it. It needs functionality to allow stopping everything when a match has been found as well as stopping further recursion down a branch if it can be determined that that branch contains nothing of interest. Thus it needs to handle {Continue, StopAll, NoRecurse} on each visit.

What was the purpose of the Python script by the way?

kraigher:
Also maybe it is better to start with having only a single Vistor per pass as it is simpler. In the future we can add a wrapper that makes a Visitor out of a Vec of Visitors.

kraigher:
Also I think it might be a bit overkill to implement a visitor function for every small item in the AST. I think maybe stopping at a reasonable granularity is better such at the statement and declaration level. 

Maybe it would be better to factor the Visitors into several types such as:
1. A Visitor going to Design Units
2. A Visitor going to Declarations & Statements
3. A Visitor going to individual terms in Names & Expressions

In that way it is easier to select how far into the details you want to recurse.

Bochlin:
The reason for almost every single item having a visitor is that the code is generated from the python script (feels a little bit like heresy). Idea was to generate the base implementation and then manually clean out the lower level parts and maintain it manually as I don't expect the AST do change dramatically.

PR #30, opened 2019-01-30
WIP: Parse comments

benreynwar:
This is a rework of the tokenizing of comments.

The 'Tokenizer' now pops LeadingComment and TrailingComment tokens, rather than incorporating the comments into the other tokens.

The 'TokenStream' skips over the token comments during normal operation so that users of TokenStream are not effected by this change.
The 'TokenStream' has new methods of "leading_comments", "trailing_comment" and "unhandled_comments" which allows parsing code access to the comments when required.


kraigher:
Sorry for the late response, life got in the way. Anyway I have had a look at this and would be ready to merge it. I agree it is a good starting point to add comments.

Issue #45, opened 2019-12-06
Add VHDL release packaging


kraigher:
In summary this still needs to be done;
- [x] Build `vhdl_parser` as packaged
- [ ] Publish github release on `v.*.*.*` tag
- [ ] Use release notes from `release_notes/v*.*.*.md` file

umarcor:
Currently, it is not straighforward to publish releases assets/artifacts from multiple GitHub Actions jobs (there is not *official* releaser as in other services). However, I believe that Rust has built-in cross-compilation (as golang), doesn't it? If so, implementing the release should be easy.

Regarding the VSCode extension, ghdl's ls and docker can be used as an example to build and distribute a VSIX file.

About docker images. On the one hand, the VSIX can be distributed in it, to allow users to attach to the container with VSCode's 'Remote - Containers' extension. On the other hand, the registry to be used needs to be discussed. GitHub Package Registry is built-in, and no separate credentials need to be managed, but it is not public: a GitHub token is required to access the images. hub.docker.com is public: anyone can `docker pull` without a token.

Xcodo:
I started looking at this in March when I was looking to get my team using Rust HDL with VSCode on our Xilinx FPGA projects on Windows. Then I had a baby and everything got shelved!

I resurected my fork and CI build last night and this morning - take a look:
GitHub - https://github.com/Xcodo/rust_hdl
TravisCI - https://travis-ci.org/Xcodo/rust_hdl

Currently the Windows build fail (new issue for that coming), but the build from before I merged the last eight month's changed ([build 13](https://travis-ci.org/Xcodo/rust_hdl/builds/621221590)) worked, so I think the cross-compilation configuration is good.

P.S. This is my first look at Rust, so I may be doing it wrong!

Xcodo:
I guess the best way would be to add a bash script or two that are called in the Travis YAML, rather than adding lots to the YAML `script` section. We might also need to build for all the environments within the same build rather than using the Travis `matrix` option to be able to do a single push to GitHub.

Happy to have my fork used as a playground for CI stuff, I can add other owners if someone else picks up the work adding the libraries or pushing releases to GitHub. I'm not going to be able to do much more for a few weeks.

kraigher:
@Xcodo good that you are already on it. Congratulation to your baby!

Xcodo:
[...]

PR #50, opened 2019-12-24
Add CI artifacts

Xcodo:
Adds artifact publishing to build.
Adds standard and IEEE libraries and packaged them with language server builds.
Working towards #45 (tagged releases remaining).


kraigher:
Looks good, it misses the removal of ieee and std libraries from example project.

kraigher:
When moving the standard libraries this code needs to be changed: https://github.com/kraigher/rust_hdl/blob/master/vhdl_parser/src/analysis/tests/util.rs#L90-L108
A test utility points out the standard libraries using a relative path, when they move you need to change this path.

Xcodo:
Sorry about that. I'll remember to actually run the tests next time!

PR #57, opened 2020-01-08
Add matching case

Bochlin:
Work in progress to add matching case statement.

Added tokenization of '?' and updated parsing of case statement to support matching case.
```
case? expr is
    when "--00" => null;
    when others => null;
end case?;
```

This update removes the error thrown by the parser but does not store in the AST that it is a matching case statement. My suggestion is to add a `is_matching: bool` element to the `ast::Selection` struct.

@kraigher Any comments?

### DIFF
--- a/vhdl_lang/src/analysis/sequential.rs
+++ b/vhdl_lang/src/analysis/sequential.rs
@@ -107,7 +107,8 @@ impl<'a> AnalyzeContext<'a> {
                 }
             }
             SequentialStatement::Case(ref mut case_stmt) => {
-                let Selection {
+                let CaseStatement {
+                    is_matching: _,
                     expression,
                     alternatives,
                 } = case_stmt;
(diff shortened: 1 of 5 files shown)


kraigher:
Nice! 

1. I think we should add a field to the AST to know that we parse it.
I do not think it should be on Selection since it is a general structure used for with-select-statements and case-generate statements. I think it is better to make CaseStatement a struct instead of a type alias of Selection with a field for the selection and one field for the is_matching

2. Either check or at least add a @TODO for checking that the end case keyword is also ?

Bochlin:
1. with-select can also be matching, it's only case-generate that is not. Intention is to do matching with-select as well. But I can start with a CaseStatement struct with an is_matching field.
2. That is checked  on line 169.

Bochlin:
@kraigher I think that this is ready to be merged now.

PR #59, opened 2020-01-10
Allow publish actions only from main repo

Xcodo:
Currently `Docker` action fails on pushing to forked repos (`Crates.io` would also fail on a publishig a tag). This adds a check on the actions so that they only act on the main repo to stop actions failing in forks.

### DIFF
--- a/.github/workflows/crates-io.yml
+++ b/.github/workflows/crates-io.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   publish:
     name: Publish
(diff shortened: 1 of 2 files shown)


kraigher:
I had to revert it because the build failed due to invalid docker file.

Xcodo:
Sorry about that. I see why it was okay on my test (it's not `master`). I'll try again. I hate CI sometimes!

PR #61, opened 2020-01-12
Add block_header and analyze guard_condition

Bochlin:
Updates for `block_statement`:
- Parsing and analysis of `block_header`
- Fix error when optional `is` is used after `guard_condition`
- Analysis of `guard_condition`

### DIFF
--- a/vhdl_lang/src/analysis/tests/resolves_names.rs
+++ b/vhdl_lang/src/analysis/tests/resolves_names.rs
@@ -1046,3 +1046,33 @@ end architecture;
 ",
     );
 }
+
+#[test]
+fn block_names_are_visible() {
+    check_code_with_no_diagnostics(
+        "
+entity ent is
+  port (ent_in : integer);
+end entity;
+
+architecture a of ent is
+  signal sig : integer;
+begin
+  blk: block (ent_in = 1) is
+    generic( gen : integer := 0 );
+    generic map ( gen => 1);
+    port( 
+      prt_in : in integer := 0;
+      prt_out : out integer := 0
+    );
+    port map (
+      prt_in => ent_in + sig,
+      prt_out => open
+    );
+  begin
+    prt_out <= gen + prt_in + sig;
+  end block;
+end architecture;
+",
+    );
+}
--- a/vhdl_lang/src/syntax/concurrent_statement.rs
+++ b/vhdl_lang/src/syntax/concurrent_statement.rs
@@ -51,7 +51,7 @@ pub fn parse_block_statement(
     stream.expect_kind(SemiColon)?;
     Ok(BlockStatement {
         guard_condition,
-        header: header,
+        header,
         decl,
         statements,
     })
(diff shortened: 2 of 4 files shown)


vhdl_lang/src/analysis/tests/resolves_names.rs
```diff
@@ -1046,3 +1046,33 @@ end architecture;
...
+    port( 
+      prt_in : in integer := 0;
+      prt_out : out integer := 0
+    );
+    port map (
+      prt_in => ent_in + sig,
+      prt_out => open
+    );
+  begin
+    prt_out <= gen + prt_in + sig;
```
kraigher: Are these types of blocks allowed to see the outside signals such as `sig` in this case?

vhdl_lang/src/syntax/concurrent_statement.rs
```diff
@@ -52,11 +51,114 @@ pub fn parse_block_statement(
...
         decl,
         statements,
     })
 }
 
+fn parse_block_header(
+    stream: &mut TokenStream,
+    diagnostics: &mut dyn DiagnosticHandler,
+) -> ParseResult<BlockHeader> {
+    let mut has_port_clause = false;
```
kraigher: Is not has_X redundant with X.is_some()?

vhdl_lang/src/analysis/tests/resolves_names.rs
Bochlin: LRM 11.2 states 

> If a block header appears in a block statement, it explicitly identifies certain values or signals that are to be imported from the enclosing environment into the block and associated with formal generics or ports.

Which I would interpret as `sig` not being visible but it's not 100% clear. ModelSim will compile the snippet without errors.

vhdl_lang/src/syntax/concurrent_statement.rs
Bochlin: They are, will update

vhdl_lang/src/analysis/tests/resolves_names.rs
kraigher: What does GHDL do? I suspect this corner of the standard is very rarely visited so there might be bugs.

Bochlin: I tried it with GHDL and some other simulators/editors as well, in all of them `sig` is visible. To me it makes more sense that adding a block header would hide all objects external to the block, both from interpreting the standard and in a general sense. The VITAL standard requires models to be structured into a set of blocks and I think that these blocks must use generics and ports.   
I would be fine with allowing `sig` to be visible as that seems to be inline with most other tools as a first step. Not sure how much work it is to hide the outer scope.

kraigher: Yes to me it would also seem to make more sense to hide the enclosing scope when using generics. Then a block with generics would be equivalent with an inlined entity instantiation. I do not think this code is very common though.

PR #62, opened 2020-01-12
Allow multiple '^' in relative external name

Bochlin:
Fixes parsing error when multiple "up-levels" are present in external name e.g.  
`<< signal ^.^.^.dut.gen(0) : std_logic >>`  

The number of up-levels are not stored and a TODO comment is added. The number up up-levels are required if elaboration checks are implemented as the hierarchy of the design must be known.

### DIFF
--- a/vhdl_lang/src/syntax/names.rs
+++ b/vhdl_lang/src/syntax/names.rs
@@ -259,19 +259,20 @@ fn parse_inner_external_name(stream: &mut TokenStream) -> ParseResult<ExternalNa
             WithPos::from(ExternalPath::Absolute(path_name), path_pos)
         },
         Circ => {
-            // @TODO Store number of '^' ("up-levels")
             stream.expect_kind(Dot)?;
+            let mut up_levels = 1;
             while stream.skip_if_kind(Circ)? {
                 stream.expect_kind(Dot)?;
+                up_levels += 1;
             }
             let path_name = parse_name(stream)?;
             let path_pos = path_name.pos.clone().combine_into(&token);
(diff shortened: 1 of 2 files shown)


vhdl_lang/src/syntax/names.rs
```diff
@@ -259,7 +259,11 @@ fn parse_inner_external_name(stream: &mut TokenStream) -> ParseResult<ExternalNa
             WithPos::from(ExternalPath::Absolute(path_name), path_pos)
         },
         Circ => {
+            // @TODO Store number of '^' ("up-levels")
```
kraigher: Why not just add an usize field and avoid the TODO?

Bochlin: Question is where to put the field then:
- `ast::ExternalName`: Field would only be valid if `path` field is relative
- `ast::ExternalPath::Relative(WithPos<Name>, u8),: Not sure if I like an unnamed u8 as it not perfectly clear what the purpose of it is.
- New struct `ast::RelativePath`
- Keep the circumflexes in the name

kraigher: I would just a field add to the `ExternalPath::Relative` enum variant.
FYI you can use named fields in enums as well to make an enum variant look like a struct:
https://doc.rust-lang.org/rust-by-example/custom_types/enum.html
But just adding a comment above the variant is probably enough without naming the fields. 

Also I would not make it u8 so it will fail with more than 255 parent levels, there is no memory footprint win to just making it an usize since external names are not so common.

PR #64, opened 2020-01-14
GitHub release process

### LINKED ISSUE #45: Add VHDL release packaging

Xcodo:
I finally have a release process that works! There's an example release created by the same YAML file (in a different branch) at [Xcodo/rust_hdl/releases/tag/v0.0.0](https://github.com/Xcodo/rust_hdl/releases/tag/v0.0.0).

I'm not totally happy, but it turns out writing a CI process by snatching half an hour here and there is very inefficient so probably best for other people to make suggestions! @kraigher @Bochlin? Also happy to explain what's going on and add comments if we need it.

Things I don't like

- The workflow runs twice when you push - once for the commit to the branch and once for the tag.
- The tag workflow builds against MSRV when it doesn't need to. Could add more conditionals to the build steps or have a separate release workflow.
- There's a lot of uploading and downloading going on. Could be fixed with a separate release workflow that starts the release then uploads the artifacts as it goes.
- The version filter doesn't actually check for semver, just tags starting with "v".

Things I do like

- **It works!**
- There is only one workflow so the build is identical to other commits and there's no danger of forgetting to make changes in two workflows.
- There is version checking against the actual binaries.

Closes #45.

### DIFF
--- a/.github/workflows/build-test-all.yml
+++ b/.github/workflows/build-test-all.yml
@@ -4,6 +4,15 @@ on: [push, pull_request]
 
 jobs:
 
+  release-notice:
+    name: This is a release build
+    if: startsWith(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Show version
+        run: echo "This is a release build for v${GITHUB_REF/refs\/tags\/v/}"
+
   build:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


kraigher:
Thank you, this is a big step towards creating a direct value stream from code to the end user.

Xcodo:
Looks like the first release worked which is great.

@Bochlin the latest release should now be available at https://github.com/kraigher/rust_hdl/releases/latest which forwards to the release page with the version in the URL.

PR #66, opened 2020-01-17
Update docker build to use built-in libraries

### LINKED ISSUE #65: Atom client requires vhdl_ls.toml
I use rust_hdl from docker and after some update language server stopped working in Atom. I was using it without any vhdl_ls.tom for simple syntax checking. Now it seems toml project file is necessary?
Zrzut ekranu z 2020-01-17 13-12-17

Xcodo:
Fixes #65.

This is untested as I don't have Docker on this machine. @mbrobbel are you able to review and test this?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,4 +3,3 @@
 Dockerfile
 example_project
 target
-vhdl_libraries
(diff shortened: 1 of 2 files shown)


mbrobbel:
Looks good to me 👍

PR #74, opened 2020-04-22
WIP: Outline concept ("textDocument/documentSymbol")

Bochlin:
I started to play around with generating an outline by implementing the ["textDocument/documentSymbol"](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_documentSymbol) request.

In order to nicely generate the hierarchical outline in the image below, the language server needs to provide the `range` of the symbol (e.g for an entity the range from the `entity` keyword until the semicolon at the end of the entity) and the `selection_range` which could be the range of the identifier (can include `entity`).

image

Creating the response is fairly straight-forward (vhdl_ls/src/document_symbol.rs) except that the full range of the AST objects are not captured by the parser. I simpler form of outline could be created by only outlining named entities and using the range of the identifier, but in that case it is not possible to build the hierarchy.

The way I see it, the range annotations in the AST should be done down to and including `block_declarative_item` and `concurrent_statement` to build a really good outline. The range annotations could also be useful for getting snippets from the code for use in e.g. hovers.

I guess the annotation could be done in two ways, either wrapping the AST struct in a WithPos<> or adding a field to the struct. Both options would require some work since it impacts both the parser and analyzer plus requiring updates to the tests. Wrapping in WithPos<> I believe would minimize the test impact but instead make traversal less ergonomic as you would have to go though WithPos<>.item. I'm not sure how big hit it would be to performance to add the annotations.

When playing around I added a `pos: Option<SrcPos>` field as that was the quickest way to get a working concept to explore.

@kraigher Do you think that this is something worth discussing further?

### DIFF
--- a/vhdl_lang/src/analysis/design_unit.rs
+++ b/vhdl_lang/src/analysis/design_unit.rs
@@ -462,10 +462,10 @@ impl<'a> AnalyzeContext<'a> {
     fn analyze_context_clause(
         &self,
         region: &mut Region<'_>,
-        context_clause: &mut [WithPos<ContextItem>],
+        context_clause: &mut ContextClause,
         diagnostics: &mut dyn DiagnosticHandler,
     ) -> FatalNullResult {
-        for context_item in context_clause.iter_mut() {
+        for context_item in context_clause.items.iter_mut() {
             match context_item.item {
                 ContextItem::Library(LibraryClause { ref name_list }) => {
                     for library_name in name_list.iter() {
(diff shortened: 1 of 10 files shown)


kraigher:
Adding a field makes sense. Maybe the name pos is a bit generic though. Maybe call it source_range or source_span.

Bochlin:
source_range sound better to me, it maps better to the use of "range" in LSP.

Closing this pull request and open a new when there is a cleaner implementation.

PR #98, opened 2020-08-23
Reload design when workspace configuration changes

### LINKED ISSUE #41: LSP should handle config file changed and/or created after startup
Currently Rust HDL looks for the config file when the LSP is started, so any changes to it (for example adding files, fixing libraries, etc) require the server/editor to be restarted.

Ideally, I think restarting or refreshing the server should happen transparently to the user, so that the process of changing and fixing the config file is faster.

HenningHolmDE:
If supported by the client, the server registers for changes to the configuration file using client/registerCapability.
When the configuration file changes, the client send a notification workspace/didChangeWatchedFiles to the server. The configuration is then loaded again and project configuration is updated:
The design state is reset, new files are added and parsed. As required for incremental document updates, source files that are already loaded are parsed from in-memory sources.

This closes #41

### DIFF
--- a/vhdl_lang/src/project.rs
+++ b/vhdl_lang/src/project.rs
@@ -31,20 +31,68 @@ impl Project {
         }
     }
 
+    /// Create instance from given configuration.
+    /// Files referred by configuration are parsed into corresponding libraries.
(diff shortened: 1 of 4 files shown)


Xcodo:
Looks good on VSCode on Windows (but you knew that!). Linux will need to wait until tomorrow.

PR #100, opened 2020-09-07
Add explicit "pure" to the syntax parser

### LINKED ISSUE #92: "pure" before function declaration/definition is flagged as unexpected
With the VHDL LS plugin (0.3.1) of VisualStudioCode (1.47.3) the VHDL keyword _pure_ is flagged as unexpected before the _function_ keyword. A list of expected keywords is shown. The keyword _impure_ is among the ones suggest, which I consider mildly ironic.
The keyword _pure_ is optional but correct in this position (the only position where _pure_ can actually appear at all I guess) and should not be flagged.

pure_function_vhdl-ls

According to the plugin's description "As VHDL LS uses the Rust HDL language server, issues related to syntax and semantic checks should be reported directly to Rust HDL".

HenningHolmDE:
The "pure" keyword was previously added to the tokenizer only.
Extending the syntax parser with "Pure" where "Impure" was used allows for parsing explicit pure functions correctly.

This closes #92.


HenningHolmDE:
@kraigher Thanks for merging!

PR #105, opened 2021-03-05
Fix short-circuiting of searches with search_with_pos.

j-zeppenfeld:
There were a few places in the AST search code that used the following pattern:
```rust
return_if_found!(searcher.search_with_pos(&self.pos).or_not_found());
```

However, `search_with_pos` returns `NotFinished` to continue searching of the region, or `Finished(NotFound)` if the position does not match. Both cases result in `NotFound`, so the above statement has no effect.

I assume this was meant to abort the search if `self.pos` does not include the position being searched for, as it does in other parts of the search code using the following pattern:
```rust
searcher.search_with_pos(&self.pos).or_else(|| { ...
```

I believe this (actually useful) pattern should be applied to the problematic cases as well - ergo this PR.

### DIFF
--- a/vhdl_lang/src/ast/search.rs
+++ b/vhdl_lang/src/ast/search.rs
@@ -546,8 +546,9 @@ fn search_pos_name(pos: &SrcPos, name: &Name, searcher: &mut impl Searcher) -> S
 
 impl Search for WithPos<Name> {
     fn search(&self, searcher: &mut impl Searcher) -> SearchResult {
-        return_if_found!(searcher.search_with_pos(&self.pos).or_not_found());
-        search_pos_name(&self.pos, &self.item, searcher)
+        searcher
+            .search_with_pos(&self.pos)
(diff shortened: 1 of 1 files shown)


kraigher:
I this fixing a bug? In that case it would be good to include a test case. 
Or is it just an equivalent refactoring? Too me it seems equivalent but I might be missing something. It becomes a matter of style if you prefer nesting continuation functions or returning early. The point of the macros was to avoid having to nest everything and make the code read more linearly.

j-zeppenfeld:
I'd call it a performance bug... `return_if_found` should be `return_if_not_found`, and `or_not_found` should be `or_found`. Both of those don't exist, so I used the nested continuation functions that do.

As it is, the statement can never return, so actually wastes time rather than speeding the search up.

j-zeppenfeld:
Alternatively, `return_if_finished` should be used.

PR #108, opened 2021-03-29
Use submodules for example_project

Xcodo:
I've left them at the same commit as `setup.sh` script currently references.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,12 @@
+[submodule "example_project/UVVM"]
+	path = example_project/UVVM
+	url = https://github.com/UVVM/UVVM.git
+[submodule "example_project/vunit"]
+	path = example_project/vunit
+	url = https://github.com/VUnit/vunit.git
+[submodule "example_project/PoC"]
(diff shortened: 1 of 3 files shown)


umarcor:
I think you need to updated the CI workflows:

- https://github.com/VHDL-LS/rust_hdl/blob/master/.github/workflows/build-test-all.yml#L39
- https://github.com/VHDL-LS/rust_hdl/blob/master/.github/workflows/build-test-all.yml#L102
- https://github.com/VHDL-LS/rust_hdl/blob/master/.github/workflows/docker.yml#L21
- https://github.com/VHDL-LS/rust_hdl/blob/master/.github/workflows/crates-io.yml#L19

Use the following instead:

```yaml
    - uses: actions/checkout@v2
      with:
        submodules: recursive
        fetch-dept: 0
```

Xcodo:
Looks to me like the default is to not init the submodules, which is what we want. The example projects are not used in the build or test process.

umarcor:
You are correct. My bad.

Issue #136, opened 2022-10-20
Create possibility to define the location of installed `vhdl_libraries`


delafthi:
Currently, `vhdl_libraries` can only be at two positions relative to the binary:

1. `../../vhdl_libraries`
2. `../vhdl_libraries` if feature `packaged` is set

It would be nice to provide the path as an argument to the `cargo install` command. However, I'm new to rust and don't know of a solution to achieve this behaviour. As such, it would be possible to use a buildscript to adjust the path and move the libraries to the defined location. Consequently, peoples would be less confused about the missing installed libraries (#113).

Currently, the sources have to be patched to change the path to wherever someone wants to install the vhdl libraries.

kraigher:
There is the VHDL_LS_CONFIG environment variable you can use:  https://github.com/VHDL-LS/rust_hdl/blob/master/vhdl_lang/src/config.rs#L287

delafthi:
I'm only talking about the installed libraries (std and ieee). Yes, `VHDL_LS_CONFIG` would be a possible solution. However, currently at https://github.com/VHDL-LS/rust_hdl/blob/master/vhdl_lang/src/config.rs#L236 the ls panics when it doesn't find the installed libraries. Conseqently, we would need to simply return when we add the installed libraries with `VHDL_LS_CONFIG`. Additionally, users would always need to define the `VHDL_LS_CONFIG` env variable for installed libraries, which isn't the best solution in my opinion.

Maybe the current patch I use for the AUR package describes better what I intend to achieve:

```patch
diff --git a/vhdl_lang/Cargo.toml b/vhdl_lang/Cargo.toml
index 166724f..4dbbff8 100644
--- a/vhdl_lang/Cargo.toml
+++ b/vhdl_lang/Cargo.toml
@@ -33,3 +33,4 @@ assert_matches = "^1"
 [features]
 default = []
 packaged = []
+installed = []
diff --git a/vhdl_lang/src/config.rs b/vhdl_lang/src/config.rs
index be9076a..6b61ed3 100644
--- a/vhdl_lang/src/config.rs
+++ b/vhdl_lang/src/config.rs
@@ -215,7 +215,9 @@ impl Config {
         let config_relative_path = if cfg!(feature = "packaged") {
             // relative to packaged bin folder
             "../vhdl_libraries"
-        } else {
+        } else if cfg!(feature = "installed") {
+            "/usr/lib/rust_hdl/vhdl_libraries"
+        }else {
[...]

PR #143, opened 2022-12-13
Search the installed libraries in a hardcoded set of locations

delafthi:
As discussed in #136, I removed the "packaged" feature and instead search in hardcoded places for the installed vhdl libraries. The current list looks like that:

```rust
 let search_paths = [
            "/usr/lib/rust_hdl/vhdl_libraries",
            "../vhdl_libraries",
            "../../vhdl_libraries",
        ];
```

Since I'm new to rust I would appreciate some comments on the code.

### DIFF
--- a/.github/workflows/build-test-all.yml
+++ b/.github/workflows/build-test-all.yml
@@ -49,13 +49,13 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --manifest-path ${{ matrix.crate }}/Cargo.toml --release --target ${{ matrix.target }} --features "packaged"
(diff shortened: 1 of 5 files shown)


kraigher:
The code looks good but I think the priority of the folders should be changed since local configuration should always take precedence over global. For example if someone were to install your linux package on their machine they could not develop rust hdl without the development build using the `/usr/lib/` libraries.

kraigher:
```rust
// Prio from most local to most global 
let search_paths = [
            "../vhdl_libraries",
            "../../vhdl_libraries",
            "/usr/lib/rust_hdl/vhdl_libraries",
        ];
```

delafthi:
Applied both of your suggestions to the code in 5b4028b78c9c3dc5a28f9947e451625cbe648f28.

PR #173, opened 2023-05-28
Analyze bit strings

Schottkyc137:
Analyze `BitString`s and report when they are not setup correctly.

```vhdl
constant v: bit_vector := 2SX""; -- Cannot expand an empty signed bit string
constant w: bit_vector := X"10G"; -- type 'BIT' does not define character 'G'
constant x: bit_vector := D"1AFFE"; -- Illegal digit 'A' for base 10
constant z: bit_vector := 8SX"0FF"; -- Truncating to 8 bit would loose information
```

I want to start collaborating on this project as I am currently using and enjoying it. This project fills an important gap for us HDL developers, so thank you for making this Open Source.

### DIFF
--- a/vhdl_lang/src/analysis/static_expression.rs
+++ b/vhdl_lang/src/analysis/static_expression.rs
@@ -4,10 +4,10 @@ use crate::Latin1String;
 use itertools::Itertools;
 use std::iter;
 
-macro_rules! byte_is_odd_decimal {
-    ($byte: expr) => {
-        ($byte - b'0') % 2 == 1
-    };
+/// returns whether `byte` is  an odd number when interpreted as decimal.
+/// byte must be between '0' and '9', but it is up to the caller to enforce this.
+fn byte_is_odd_decimal(byte: u8) -> bool {
+    (byte - b'0') % 2 == 1
 }
 
 /// Converts a decimal string (i.e. "123") to a binary string (i.e. "1111011").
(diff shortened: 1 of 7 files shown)


kraigher:
Are you sure about the warning of "Integer too large for 64-bit unsigned"? I think bitstrings can define vectors of any length.

kraigher:
Also remember to run `cargo fmt`

Schottkyc137:
> Are you sure about the warning of "Integer too large for 64-bit unsigned"? I think bitstrings can define vectors of any length.

You're right (at least I haven't seen any limitation in the reference). I have changed the implementation accordingly (it now converts arbitrary decimal strings to binary strings, regardless of their length)

vhdl_lang/src/analysis/static_expression.rs
```diff
@@ -204,18 +208,26 @@ pub(crate) fn bit_string_to_string(
...
-    // For example, 2SB""
+    // For empty signed bit-strings it is unclear what the expanded value should be,
+    // according to the reference:
+    // For example, 2SB"" could be
     // 1) A string containing '0's, i.e. "00"
     // 2) An error
     // According to the standard, the padding value should be the leftmost character in the string
     // but an empty string does not have a leftmost character.
     if simplified_value.len() == 0 {
+        println!("");
```
kraigher, line 219: Remove leftover debug printout

vhdl_lang/src/analysis/static_expression.rs
```diff
@@ -0,0 +1,478 @@
+use crate::analysis::static_expression::BitStringConversionError::EmptySignedExpansion;
+use crate::ast::{BaseSpecifier, BitString};
+use crate::Latin1String;
+use itertools::Itertools;
+use std::iter;
+
+macro_rules! byte_is_odd_decimal {
```
kraigher, line 7: This could just be a function, I typically only use macros when a function is not possible.

kraigher:
I wrote two comments. As you can see above the "cargo clippy" check that I run as a CI check had some complaints as well.

PR #175, opened 2023-06-03
Enable initial support for static expressions

Schottkyc137:
I want to enable support for static expression so that the following couple of examples will be recognized as errors:
```VHDL
constant x: bit_vector(2 down 0) := "0011"; -- Error: Left hand side of assignment has size 3 but right hand side has size 4
constant WIDTH: natural := 5;
constant y: bit_vector(WIDTH - 1 down 0) := "01" & "10"; -- Same as above
```

There are many more places where this is needed to recognize some more complex errors, some common examples include:
- Checking the bit-width for ports and generics in instantiated components
- Check full overlap for `case` statements
- Check the range for range constraint integers
- Check that something that expects a static value (i.e. constants) also get a static value

As this will probably be a fairly big change in the codebase I want to make sure the implementation makes sense from the very beginning. My idea is as follows:
- When analyzing expressions, return the static value of that expression. 
  - When the expression is not static, return the special case `StaticValue::NonStatic`.
  -  If there is another error, return `StaticValue::HasOtherError`. 
  - As this might be a longer process, return `StaticValue::Unimplemented` to retain the status where this implementation doesn't change anything
- Assignments, case statements and others can then be analyzed using the obtained static value.

Any comments on this approach? Do you think this will work or do you have something else in mind? I'm happy to discuss further approaches.

### DIFF
--- a/vhdl_lang/src/analysis/assignment.rs
+++ b/vhdl_lang/src/analysis/assignment.rs
@@ -9,6 +9,7 @@ use super::analyze::*;
 use super::named_entity::*;
 use super::region::*;
 use super::target::AssignmentType;
+use crate::analysis::static_expression::StaticValue;
 use crate::ast::*;
 use crate::data::*;
 
@@ -127,12 +128,12 @@ impl<'a> AnalyzeContext<'a> {
         ttyp: Option<TypeEnt<'a>>,
         expr: &mut WithPos<Expression>,
         diagnostics: &mut dyn DiagnosticHandler,
-    ) -> FatalResult {
+    ) -> FatalResult<StaticValue> {
         if let Some(ttyp) = ttyp {
-            self.expr_with_ttyp(scope, ttyp, expr, diagnostics)?;
+            self.expr_with_ttyp(scope, ttyp, expr, diagnostics)
         } else {
-            self.expr_unknown_ttyp(scope, expr, diagnostics)?;
+            self.expr_unknown_ttyp(scope, expr, diagnostics)
+                .and(Ok(StaticValue::Unimplemented))
         }
-        Ok(())
     }
 }
--- a/vhdl_lang/src/analysis/declarative.rs
+++ b/vhdl_lang/src/analysis/declarative.rs
@@ -9,6 +9,7 @@ use super::formal_region::RecordRegion;
 use super::named_entity::*;
 use super::names::*;
 use super::*;
+use crate::analysis::static_expression::{StaticConstraint, StaticValue};
 use crate::ast;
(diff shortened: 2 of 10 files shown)


kraigher:
I think it is a difficult feature to implement correctly if you look beyond the simple cases. The difficulties comes from:

1. You need to actually execute VHDL code to determine the result of expressions. Even a simply expression like `1 + 5` could use a custom function for `+` that the user wrote themselves. 

2.  Many constants such as those from generics are not statically known when the code is analyzed. The generics are only known during the elaboration phase. VHDL LS does not do any elaboration of the design today.

3. Functions can declare arrays using constants passed as parameters for the width.

Of course you could try to do something simpler that ignores the above difficulties and treats them as unknown. The implementation would have to accept that there are many cases where it cannot know the answer and that would the the `StaticValue::NonStatic` in your case.

Keep in mind that when developing VHDL LS I have one principles to never accept false diagnostic messages. The implementation must be aware of its limitations and rather remain silent than complain about things that it cannot know for sure. It is very annyong for the user to see messages that are false and makes the tool much less usefull that if it would have remained silent. This is very important when incrementally developing new lints as they will naturally be very limited in the beginning. 

This principle means in the above examples you cannot assume that `+` is the builtin `+` function but must actually check that it is not a user defined custom `+` function.

Schottkyc137:
I agree with your general considerations.

As far as scope goes I still think that if we limit ourselves to locally static primaries (as defined in LRM 9.4.2) there is still  value in having the capabilities described above. Limiting to this case still covers some of the more common cases. This would address some of your issues:

1) I would only use some pre-defined functions and not statically elaborate user-defined functions or operators
2) I also think that this is out of scope, at least for now.
3) I assume you mean something like `to_unsigned(x, width)`? Probably also out of scope though as this is a pre-defined function there might be some simple assertions that can be made.

I shall keep the principle of no false diagnostic messages in mind. However with the nice work you have done on function overload matching some pre-defined functions (again, as defined in LRM 9.4.2) could already be evaluated within the scope of analysis.

kraigher:
> 3. Functions can declare arrays using constants passed as parameters for the width.
I mean things like this:
```vhdl
function func(arg : natural) return std_logic_vector is
   variable v1 : std_logic_vector(arg-1 downto 0);
begin
   return v1;
end;
```


> However with the nice work you have done on function overload matching some pre-defined functions (again, as defined in LRM 9.4.2) could already be evaluated within the scope of analysis.
Yes the analys step will set a reference on each operator which points to where it was defined, so that can be used. You should probably add a unit where a user-defined operator is ignored.

Issue #147, opened 2023-01-20
Subtype error gives proken pkg


thoste:

kraigher:
Thanks for reporting, should be easy to fix. As I am in the middle of some other work right now but will fix it in the near future.

PR #186, opened 2023-08-11
Feat: Enable type attributes

### LINKED ISSUE #184: [Bug] Errors reported when using consecutively following attributes
**Describe the bug**
The following error(s) are reported for the snippet below: **_Type cannot be used as an expression_  ``vhdl ls``**
It seems to be the LS can't handle consecutively following attributes.

**Code**
```vhdl
entity tb
  generic (
    runner_cfg: string := runner_cfg_default;
    tb_path: string
  );
end entity;

architecture behavioural of tb is
  data_generator: process
    variable data_index: natural range 0 to 9 := 9;

    procedure report_index is
      variable data_index_1: natural range 0 to data_index'subtype'high;
      variable data_index_0: natural range 0 to data_index'subtype'high;
    begin
      data_index_1 := data_index;
      data_index_0 := data_index - 1;

      report "data_index_1: " & to_string(data_index_1) severity note;
      report "data_index_0: " & to_string(data_index_0) severity note;
    end procedure;

    begin
      for i in 0 to data_index'subtype'high loop
        report_index;
        if data_index > data_index'subtype'low then
          data_index := data_index - 1;
        else
          data_index  := data_index'subtype'high;
        end if;
      end loop;
    end process;
end architecture;
```

**Add. info:**
- The code works with QuestaSim 2022.4:
 - OS: Windows 11
 - VSCode version: 1.80.1
 - TerosHDL Version: v5.0.6

**Screenshots**
image

**Additional context**
Is there a way how to mute the error for the mean time?
I reported this also in Teros-HDL repo - https://github.com/TerosTechnology/vscode-terosHDL/issues/429 - as well but it seems to be a bug related to this repo.

Schottkyc137:
Closes #184 

Implementation details:
Introduces a new return type `AttrResolveResult` which is either a type (for attributes such as `'subtype`, `'element`) or a value (for most other attributes such as `'high`, `'low`, ...)

Based on this attribute the name can be resolved correctly and `a'subtype'high` will resolve to the correct type while `a'subtype` as an expression yields an error

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -909,7 +909,7 @@ checksum = "711b9620af191e0cdc7468a8d14e709c3dcdb115b36f838e601583af800a370a"
 
 [[package]]
 name = "vhdl_lang"
-version = "0.64.0"
+version = "0.65.0"
 dependencies = [
  "assert_matches",
  "clap",
@@ -929,7 +929,7 @@ dependencies = [
 
 [[package]]
 name = "vhdl_ls"
-version = "0.64.0"
+version = "0.65.0"
 dependencies = [
  "clap",
  "env_logger",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -909,7 +909,7 @@ checksum = "711b9620af191e0cdc7468a8d14e709c3dcdb115b36f838e601583af800a370a"
 
 [[package]]
 name = "vhdl_lang"
-version = "0.65.0"
+version = "0.64.0"
 dependencies = [
  "assert_matches",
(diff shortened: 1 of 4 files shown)


Schottkyc137:
Note: Also related to #147 since a `type_mark` can now simply be parsed as `name` (as opposed to a `selected_name`) as `resolve_name` now correctly resolves type of a name. The only "problematic" thing is the diagnostic handling in the `AnalyzeContext` (`EvalResult` vs `AnalyzeResult`).

PR #197, opened 2023-08-29
Return \n to end of line in case it's deleted

### LINKED ISSUE #193: Emacs strange Unexpected EOF
I am using emacs lsp-mode with vhdl_ls.

Sometimes when I am writing the code, I get a strange `Unexpected EOF`. This happens only with vhdl_ls, in other languages everything works fine. I am unable to get to the bottom of it and find the actual cause. The only way to get rid of the error I found is to restart the lsp.

I have made a video showing one of the ways I get to the `Unexpected EOF` error here: https://youtu.be/xV2RFoDygac. It shows how I first get the error, and then if I make changes - including getting back to the previously working code - the error still stays there. It also shows that I cannot get to the definition of that entity. I suppose that has to do with the inner workings of the library. When I restart the workspace, it starts working again. 

If anyone has any idea where the error could come from or any help regarding how to investigate further to get to the underlying issue, I would be glad.

I am also getting an error when restarting the workspace, but I will create a separate issue for that.

Rutherther:
One possible way to resolve #193.

It looks to me like LSP really does delete \n at the end of the line in the did change document event, if that's true, this seems to me like a plausible solution.

It's possible I am overlooking something, in that case this may be just a hot fix ignoring the actual cause.

### DIFF
--- a/vhdl_lang/src/data/contents.rs
+++ b/vhdl_lang/src/data/contents.rs
@@ -115,6 +115,12 @@ impl Contents {
             }
         }
 
+        // New line should always be at the end of line, in case
+        // it's deleted by accident somehow, return it back.
+        if merged_content.chars().last().unwrap_or('\0') != '\n' {
+            merged_content.push('\n');
+        }
+
         let end_line = std::cmp::min(self.lines.len().saturating_sub(1), end_line);
         self.lines
             .splice(


kraigher:
I still do not understand if it is a bug in the server or the client that sends incorrect change messages. Is it possible to reproduce the problem in a unit test? That would help understand the scenario.

Rutherther:
To be honest, I am not sure about that either. What do you think the unit test should put as the change? Removal of \n at the end of a line? - That is what it looks to me the client sends. I think it would be for the best to just log the changes to a file, and I will then send this file to the linked issue. From that it should be more clear what is happening.

Even if this were a bug on the client, I would still think that since it's necessary for all lines to end with \n, it should be added anyway when it's removed, or do you think otherwise? - Of course if there is another issue causing this, that should be solved as well.

kraigher:
I think the unit test makes it a lot easier to know why the code needs to change.  
It is easier for me to review the unit test than to check the youtube videos. 
It also ensures the "corner case" will work in the future.
The files `contents.rs` you changed contains lots of unit test for the change mechanism to it should be easy to add one more. 

I suspect the issue is that the client sends a change interval that is "to the right of" the end of line.
To delete a line I think the end interval must be the next line. 
So a reproducing unit test would send a change that goes beyond the line it affects without going to the next line.

Rutherther:
> I suspect the issue is that the client sends a change interval that is "to the right of" the end of line.
> To delete a line I think the end interval must be the next line.

Yeah, I think that is the case, thanks. I will make the unit test soon.

Rutherther:
I am so sorry about that, I had and have other stuff on my mind and didn't get around to that. I will probably not be able to get to that soon

kraigher:
Well close this for now. You can always re-open it when you have time to finish it.

PR #200, opened 2023-09-23
Include tokens in ast

Schottkyc137:
As we talked about in the element group, this PR aims to implement the basic features needed to record token information in the AST.
This PR is theoretically ready to merge but should be reviewed carefully as it brings some substantial changes.
Once the review process is done, I will add some more tests and documentation.

## Concept
Each token is uniquely identified using a `TokenId`. This is a thin wrapper around a `usize`. However, it can only be constructed in the `TokenStream`. This prevents accidental accesses and makes it safe to use as index into arrays. Using the trait `TokenAccess`, types declare that they are capable of receiving a `TokenId` and return the token identified by this id. Methods of the `TokenStream` such as `pop_if_kind`, `expect_kind`, ... now return a `TokenId` instead of the `Token`. This id is then stored in the AST. When performing actions on the AST, the TokenAccess is provided so that the token values can lateran received. The tokens themselves are stored with each primary unit inside the AST.

This PR lays the groundwork for other concepts such as formatting or completion.


vhdl_lang/src/ast.rs
```diff
@@ -1082,19 +1084,45 @@ pub struct LabeledConcurrentStatement {
...
 pub struct LibraryClause {
-    pub name_list: Vec<WithRef<Ident>>,
+    pub library_token: TokenId,
+    pub name_list: IdentList,
+    pub semi_token: TokenId,
+}
+
+/// Represents a token-separated list of some generic type `T`
+#[derive(PartialEq, Debug, Clone)]
+pub struct SeparatedList<T> {
```
kraigher, line 1094: Do we really need to store the TokenId:s of the separators? Can't they be found by accessing the previous token from the list elements if needed?

kraigher:
I think it looks good. I wrote a single comment. In general I would say we probably do not need to store the `TokenId` for everything such as semi-colons in the AST as long as we have access to the token stream and can look at the next or previous token. But I do not see a big problem with doing it this way either.

vhdl_lang/src/ast.rs
Schottkyc137: Yes, if the element in the list all have the same token length. For example, for a `library` clause this works:
```
library ieee, std, work;
```
All tokens can be uniquely addressed using the `library` token. However, for more complex `SeparatedList`s this is no longer possible:
```
use ieee.std_logic_1164.all, std.env.all;
```
The count of tokens is not necessarily the same for all elements in the name-list.
My opinion is that this is an optimization for some later point when all token locations are known. Then, given a list of `Name`s, one might be able to advance by the count of tokens in that name to get the location of the `comma` token.

Schottkyc137:
> I think it looks good. I wrote a single comment. In general I would say we probably do not need to store the `TokenId` for everything such as semi-colons in the AST as long as we have access to the token stream and can look at the next or previous token. But I do not see a big problem with doing it this way either.

See my comment. I could optimize this for the `library` case as the list only contains identifiers. However, this seems like a small benefit for increased fragmentation of the code.

kraigher (approved):

vhdl_lang/src/syntax/parse_context.rs
```diff
@@ -0,0 +1,57 @@
+use crate::ast::{ContextItem, ContextReference, LibraryClause, UseClause};
```
kraigher, line 1: Add a copyright notice to new files that is the same as in other files.

vhdl_lang/src/syntax/separated_list.rs
```diff
@@ -0,0 +1,89 @@
+use crate::ast::{IdentList, NameList, SeparatedList, WithRef};
```
kraigher, line 1: Add a copyright notice to new files that is the same as in other files.

vhdl_lang/src/syntax/test.rs
```diff
@@ -39,6 +42,31 @@ pub struct CodeBuilder {
     pub symbols: Arc<Symbols>,
 }
 
+impl AnyDesignUnit {
+    pub fn entity(&self) -> &EntityDeclaration {
+        match self {
+            AnyDesignUnit::Primary(AnyPrimaryUnit::Entity(ent)) => ent,
+            _ => unreachable!("Expected entity"),
```
kraigher, line 49: I think panic! is more appropriate here. Unreachable is only used if it can be known locally. Here it is up to the caller to ensure we never call this on the from `DesignUnit`.

I would also have named the function `expect_entity` if it panics on non-entites. Otherwise I would have returned an `Option` and handled it in the caller.

vhdl_lang/src/syntax/test.rs
```diff
@@ -39,6 +42,31 @@ pub struct CodeBuilder {
...
+        }
+    }
+}
+
+impl ContextItem {
+    pub fn library_clause(&self) -> &LibraryClause {
+        match self {
+            ContextItem::Library(lib) => lib,
+            _ => unreachable!("Expected library clause"),
```
kraigher, line 58: See similar comment.

vhdl_lang/src/syntax/test.rs
```diff
@@ -39,6 +42,31 @@ pub struct CodeBuilder {
...
+        }
+    }
+
+    pub fn context_reference(&self) -> &ContextReference {
+        match self {
+            ContextItem::Context(ctx) => ctx,
+            _ => unreachable!("Expected context clause"),
```
kraigher, line 65: See similar comment.

kraigher (changes requested):

kraigher:
Just had a few comments. Mainly the copyright header is missing. It is important to avoid having the project license status be unclear. I should probably add a CI-check for it

kraigher (approved):

PR #201, opened 2023-10-08
Initial support for instantiation completions

Schottkyc137:
This PR lays the groundwork for completions for instantiations. With this PR, requesting completions for a well-formed generic or port map will list available generics or ports.

Also contains a `Visitor` implementation for easy traversal of the AST.

Things that are not considered in this PR:
- More elaborate completions, i.e. instead of completing `port_name`, the tool should be able to complete `port_name => ${0}` with an optional comma
- Completing malformed instantiations, for example empty port/generic maps, those with trailing comma, ...


kraigher:
Some quick feedback. Would it be possible to split the PR into several parts? It seems like a mix of refactoring, adding visitor and also adding completions. It is a lot of mental effort for me to review it all at once and keep track of the comments. Could you try to split it into at least 2-3 separate logical PRs with `git rebase -i`?

Regarding the Visitor, are you aware of the ´search.rs´? It might be a bit of duplication which is unfortunate. Any thought's on how they could be merged?

Schottkyc137:
> Some quick feedback. Would it be possible to split the PR into several parts? It seems like a mix of refactoring, adding visitor and also adding completions. It is a lot of mental effort for me to review it all at once and keep track of the comments. Could you try to split it into at least 2-3 separate logical PRs with `git rebase -i`?


Alright I have tried. It's not as simple as I initially thought because everything does kind of belong together.
The refactoring of the map aspect is required to store the tokens, which is required for the completion which in turn requires the visitor (which has some methods that depend on the reworked Map). I have extracted the autocompletions – but not with the rebase option, my commit history is not as clean for this as it seems :D

So now it's split into two MR's, the first being the refactoring and the second being the completion and the visitor. The second is still large but it depends on the first.

> 
> Regarding the Visitor, are you aware of the ´search.rs´? It might be a bit of duplication which is unfortunate. Any thought's on how they could be merged?

I have had a look at the `search.rs` module, yes. For my case, it doesn't quite work I think.
The main difference between the `Visitor` and the `Searcher` is that the visitor is "AST-oriented" (i.e. you operate directly on the AST) while the `Searcher` is "Concept-oriented" for a lack of a better word. 
As an example – if you want to get the `EntityDeclaration` object for a given `EntityId`, in the searcher you would search for a declaration and then see if a declaration with that Id exists and whether it is an entity. 
When you want to do that in the visitor, you first traverse all entity declarations and only then check whether the found entity declaration matches that given ID (I hope my logic makes sense here).
So I think the searcher can be re-formulated using the Visitor, but not the other way around. Essentially, this would mean replacing the `Search` trait with a single Visitor implementation.

kraigher:
Firstly, I merged the first PR so this can now be rebased.

Secondly, the `search.rs` code is kind of a Visitor but more advanced in that it searches the AST in lexical order and supports early exit. It is an overhead to implement something that traverses the entire AST so I think it is adding technical debt that we would have both a Visitor and a Search trait. I am not saying the Search trait is the best way to do it though but I would really prefer if we do not need both. In that case I would prefer a PR that just replaces the Search trait with the Visitor if it is a better solution.

Schottkyc137:
> Firstly, I merged the first PR so this can now be rebased.
> 
> Secondly, the `search.rs` code is kind of a Visitor but more advanced in that it searches the AST in lexical order and supports early exit. It is an overhead to implement something that traverses the entire AST so I think it is adding technical debt that we would have both a Visitor and a Search trait. I am not saying the Search trait is the best way to do it though but I would really prefer if we do not need both. In that case I would prefer a PR that just replaces the Search trait with the Visitor if it is a better solution.

I'm also happy to do this with the `Searcher` if the searcher can do what is needed for completion with the same ease. Is there a simple way of going through every element with a `MapAspect`, i.e. an `EntityDeclaration`, a `ComponentDeclaration` and a `PackageDeclaration` extracting the values of the ports or generics? Or how would one go about finding out whether the cursor is in an Architecture declarative part or an architecture statement part?
Happy to do either – employ the search trait here or replace the search trait with the visitor.
Concerning the latter: what exactly do you mean by 'Lexical Order'?
Concerning the early exit – the `Visitor` also supports this

kraigher:
The searcher visits every declaration so it should be possible with no or few modifications to visit every declaration with a map aspect.

kraigher:
I looked at the rebased code and it seems fine and I have no problem merging. My only issue is that I think it adds technical debt to have both a Visitor and a Searcher. Someone will have to clean that up eventually so I would prefer if you could unify it now.

Schottkyc137:
> The searcher visits every declaration so it should be possible with no or few modifications to visit every declaration with a map aspect.
> 
> https://github.com/VHDL-LS/rust_hdl/blob/f5aa115d4c770876ada9baabc5b44d8499f4a4d1/vhdl_lang/src/ast/search.rs#L43

Ah I remember another thing that was difficult: In my current implementation, the visitor needs to visit the AST twice, interleaved.
First, the `AutocompletionVisitor` looks for instantiation statements. Then, once such a statement has been found, the `PortsOrGenericsExtractor` traverses the AST to get the names of the `Formal`s. This is no problem with the `Visitor` as the nodes are all immutable (i.e. I can simply let multiple visitors run over the AST at once). However, I had difficulties with the `Searcher` because it treats the nodes as mutable, in which case I need a lock on the AST. Then, when the second searcher tries to extract the information, a deadlock situation occurred. Is there something I'm missing? Is there another way of traversing the AST while not acquiring a lock?

PR #204, opened 2023-10-15
Improve diagnostics for map aspects

Schottkyc137:
Improve the following diagnostics for map aspects:

- Empty association list (generic map or port map): Now "Association list cannot be empty"
- Trailing comma: Same error message but parsing goes on
- Extraneous commas (i.e. `port map (a => a,, b => b)`): Now "extraneous comma"

This also enables better completions as the parser can parse more semantically incorrect code.

### DIFF
--- a/vhdl_lang/src/syntax/separated_list.rs
+++ b/vhdl_lang/src/syntax/separated_list.rs
@@ -10,7 +10,6 @@ use crate::syntax::common::ParseResult;
 use crate::syntax::names::parse_name;
 use crate::syntax::Kind::Comma;
 use crate::syntax::{kind_str, Kind, Recover, TokenAccess, TokenStream};
-use std::fmt::Debug;
 
 /// Skip extraneous tokens of kind `separator`.
 /// When there are any extra tokens of that kind, mark all the positions of these tokens as erroneous
--- a/vhdl_lang/src/syntax/separated_list.rs
+++ b/vhdl_lang/src/syntax/separated_list.rs
@@ -8,9 +8,29 @@ use crate::ast::{IdentList, NameList, SeparatedList, WithRef};
 use crate::data::{DiagnosticHandler, DiagnosticResult};
 use crate::syntax::common::ParseResult;
 use crate::syntax::names::parse_name;
-use crate::syntax::Kind::{Comma};
-use crate::syntax::{Kind, TokenStream};
+use crate::syntax::Kind::Comma;
+use crate::syntax::{kind_str, Kind, TokenAccess, TokenStream};
 use std::fmt::Debug;
+
+/// Skip extraneous tokens of kind `separator`.
+/// When there are any extra tokens of that kind, mark all the positions of these tokens as erroneous
+fn skip_extraneous_tokens(
+    stream: &TokenStream,
+    separator: Kind,
+    diagnostics: &mut dyn DiagnosticHandler,
+) {
+    if let Some(separator_tok) = stream.pop_if_kind(separator) {
+        let start_pos = stream.get_pos(separator_tok);
+        let mut end_pos = start_pos;
+        while let Some(separator_tok) = stream.pop_if_kind(separator) {
+            end_pos = stream.get_pos(separator_tok)
+        }
+        diagnostics.error(
+            start_pos.combine(end_pos),
+            format!("Extraneous '{}'", kind_str(separator)),
+        );
+    }
+}
 
 /// Parses a list of the form
 ///   `element { separator element }`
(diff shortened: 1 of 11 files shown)


vhdl_lang/src/syntax/separated_list.rs
```diff
@@ -5,39 +5,71 @@
...
+        );
+    }
+}
 
 /// Parses a list of the form
 ///   `element { separator element }`
 /// where `element` is an AST element and `separator` is a token of some `ast::Kind`.
 /// The returned list retains information of the whereabouts of the separator tokens.
-pub fn parse_list_with_separator<F, T>(
+pub fn parse_list_with_separator<F, T: Debug>(
```
kraigher, line 39: `T: Debug` seems like an unrelated change leftover from debugging.

vhdl_lang/src/syntax/separated_list.rs
```diff
@@ -5,39 +5,71 @@
...
+    separator: Kind,
+    diagnostics: &mut dyn DiagnosticHandler,
+) {
+    if let Some(separator_tok) = stream.pop_if_kind(separator) {
+        let start_pos = stream.get_pos(separator_tok);
+        let mut end_pos = start_pos;
+        while let Some(separator_tok) = stream.pop_if_kind(separator) {
+            end_pos = stream.get_pos(separator_tok)
+        }
+        diagnostics.error(
```
kraigher, line 28: Did you consider creating one diagnostic per token instead?

vhdl_lang/src/syntax/separated_list.rs
```diff
@@ -5,39 +5,71 @@
...
 {
     let mut items = vec![parse_fn(stream)?];
     let mut tokens = Vec::new();
-    while let Some(separator) = stream.pop_if_kind(separator) {
-        items.push(parse_fn(stream)?);
-        tokens.push(separator);
+    while let Some(separator_tok) = stream.pop_if_kind(separator) {
+        skip_extraneous_tokens(stream, separator, diagnostics);
+        tokens.push(separator_tok);
+        match parse_fn(stream) {
```
kraigher, line 53: The first call to `parse_fun` uses the ?-operator for early exit. But inside the loop there is no longer early exit for `parse_fun` failures after this change. This is a bit inconsistent.

I do not see any test case that requires the inner loop `parse_fn` to not be early exit. Are you sure it is needed and was not an early idea that was then later not used?

kraigher (changes requested):

vhdl_lang/src/syntax/separated_list.rs
Schottkyc137: I didn't so far. This is probably an extreme edge-case so I feel indifferent to whatever solution. Probably a single diagnostic per token would have been easier to implement. Still, I think it's better (and simpler for the user) to have one single error message for all tokens.

vhdl_lang/src/syntax/separated_list.rs
Schottkyc137: This is indeed inconsistent. Do you think my change (not short-circuiting when the inner `parse_fn` fails) or the former (always short-circuiting) is better?
Personally, I think that the new change might improve completion (i.e. `a => ,` will be incorrect but completions still work). However, I'm not sure if this leads to spurious error messages. What do you think?

kraigher: I think it is ok as long as there is a test for the case where it provides value.

vhdl_lang/src/syntax/separated_list.rs
kraigher: Its not a big deal. I just think it would make the code simpler. It is not a blocking comment from me.

kraigher:
I resolved one comments and wrote an answer on another

Schottkyc137:
> I resolved one comments and wrote an answer on another

I was a bit busy last week. Should be better now.

vhdl_lang/src/syntax/separated_list.rs
```diff
@@ -47,7 +48,7 @@ where
     let mut items = vec![];
     let mut tokens = Vec::new();
     loop {
-        match parse_fn(stream) {
+        match parse_fn(stream).or_recover_until(stream, diagnostics, |kind| kind == separator) {
```
kraigher: Are you sure `or_recover_until` is a good idea here? I am worried it will cause the parsing to consume tokens far beyond the separated list. For example

´´´vhdl
a := (0 => ); -- Recover until starts here

signal this_is_skipped : boolean;

signal foo : std_logic := func(1,  2); -- Next separator is here
´´´

The recover_until "concept" is not really sound for this reason. I added it originally to skip until the next keyword in declarative parts. There needs to be a more advanced condition to stop skipping over tokens than just finding the next separator. For example also looking at semi-colon or typicall leading keywords. In the above case the ending ) should also cause the consumption of tokens to stop.

kraigher (changes requested):

vhdl_lang/src/syntax/separated_list.rs
Schottkyc137: Yea I noticed this too (fortunately there were enough unit tests to check this).
In the current solution, an optional `recover_token` can be supplied. If that token is supplied, the parser will skip until that token (or the separator token) is seen. If the token is not supplied, the parser will behave like it previously did (and skip early).

kraigher: So are you saying this is not a problem or is it something that needs to be fixed? I do not spot a recover token in the line I commented on?

Schottkyc137: I think I requested review too early. After I requested review, I pushed another commit where I changed the line that you commented. Sorry about that

kraigher (approved):

PR #205, opened 2023-10-24
More elaborate completions

Schottkyc137:
Improves the quality of completions

- Enables snippets
- Enables custom completion icons
- Lazy Documentation for symbols (Documentation only appears when hovering the completion item)
- Different label vs. text that is inserted
- Detail text for completions

Caveats:
- The documentation doesn't render in VSCode for an unknown reason
- The 'detail' text only renders after requesting (though it is in the LSP response initially).
- ~~The overload resolution for completions probably could be better and utilize the resolution that is already implemented. Just wasn't sure how to incorporate this.~~

I'm not sure how these caveats can be resolved. I'm still going to open this PR since these caveats are not breaking anything. However, if someone knows how to resolve them I'm happy to implement the better solution.

### DIFF
--- a/vhdl_lang/Cargo.toml
+++ b/vhdl_lang/Cargo.toml
@@ -25,7 +25,6 @@ parking_lot = "0"
 dunce = "1"
 pinned_vec = "0"
 itertools = "0"
-serde = { version = "1.0.156", features = ["derive"] }
 
 [dev-dependencies]
 tempfile = "3"
--- a/vhdl_lang/Cargo.toml
+++ b/vhdl_lang/Cargo.toml
@@ -25,6 +25,7 @@ parking_lot = "0"
 dunce = "1"
 pinned_vec = "0"
 itertools = "0"
+serde = { version = "1.0.156", features = ["derive"] }
 
 [dev-dependencies]
 tempfile = "3"
--- a/vhdl_lang/src/analysis/completion.rs
+++ b/vhdl_lang/src/analysis/completion.rs
@@ -3,7 +3,8 @@ use crate::ast::visitor::{Visitor, VisitorResult};
 use crate::ast::{
     AnyDesignUnit, AnyPrimaryUnit, AnySecondaryUnit, ComponentDeclaration, Declaration,
     EntityDeclaration, HasUnitId, InstantiationStatement, InterfaceDeclaration, MapAspect,
-    PackageDeclaration, SubprogramDeclaration, UnitKey, WithDecl, WithRef,
+    ObjectClass, PackageDeclaration, SubprogramDeclaration, SubprogramDesignator, UnitKey,
+    WithDecl, WithRef,
 };
 use crate::data::{ContentReader, Symbol};
 use crate::syntax::Kind::*;
@@ -15,6 +16,15 @@ use std::default::Default;
 
 pub enum CompletionKind {
     Module,
+    Constant,
+    Signal,
+    Variable,
+    File,
+    Type,
+    Function,
+    Operator,
+    Attribute,
+    Alias,
 }
 
 pub enum CompletionItemMode {
@@ -48,7 +58,11 @@ impl CompletionItem {
         }
     }
 
-    pub fn from_decl<T>(decl: &WithDecl<T>, root: &DesignRoot) -> CompletionItem
+    pub fn from_decl<T>(
+        decl: &WithDecl<T>,
+        root: &DesignRoot,
+        kind: CompletionKind,
+    ) -> CompletionItem
     where
         T: ToString,
     {
@@ -57,11 +71,11 @@ impl CompletionItem {
             decl.decl
                 .map(|id| root.get_ent(id).describe())
                 .unwrap_or_default(),
-            CompletionKind::Module,
+            kind,
         )
     }
 
-    pub fn from_ref<T>(decl: &WithRef<T>, root: &DesignRoot) -> CompletionItem
+    pub fn from_ref<T>(decl: &WithRef<T>, root: &DesignRoot, kind: CompletionKind) -> CompletionItem
     where
         T: ToString,
     {
@@ -70,7 +84,7 @@ impl CompletionItem {
             decl.reference
                 .map(|id| root.get_ent(id).describe())
                 .unwrap_or_default(),
-            CompletionKind::Module,
+            kind,
         )
     }
 }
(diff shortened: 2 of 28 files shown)


kraigher:
I am busy until next week. So expect a review in the middle of next week.

vhdl_lang/src/analysis/completion.rs
```diff
@@ -2,17 +2,122 @@ use crate::analysis::DesignRoot;
...
+    File,
+    Type,
+    Function,
+    Operator,
+    Attribute,
+    Alias,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum CompletionItemMode {
```
kraigher, line 34: This just seems like a duplicate of the lsp_types. Maybe the completion.rs file should not really be in the vhdl_lang crate but rather in the vhdl_ls crate and use the lsp_types directly. Completion is mostly a language server concern anyway.

vhdl_lang/src/analysis/named_entity/arena.rs
```diff
@@ -39,9 +40,9 @@ pub struct LocalId(u32);
 /// Arena allocators used to store named entities
 
 /// Local arena used for single design unit in a separate thread
-struct LocalArena {
+pub struct LocalArena {
```
kraigher, line 43: Where these changes really needed? I cannot find them used anywhere.

vhdl_lang/src/analysis/named_entity/arena.rs
```diff
@@ -120,6 +121,10 @@ impl<'a> FinalArena {
     pub fn clear(&mut self) {
         self.refs.clear();
     }
+
+    pub fn items(&self) -> impl Iterator<Item = &Arc<LocalArena>> {
```
kraigher, line 125: Do we really need to iterate over LocalArenas? I would prefer if it was private. Maybe add a function to iterate over all entities?

Schottkyc137: Turns out this was superfluous as well :D I removed the function altogether

vhdl_lang/src/analysis/completion.rs
Schottkyc137: I generally agree. I'll see if it's simple to migrate.
The initial reason I put this into the `vhdl_lang` crate is because of the API design of `vhdl_ls`, which lacks some public features needed for completion. But maybe that has changed / is simpler than I initially thought.
Anyhow, it's probably a good idea to refactor this in the mid / long term (or at least before adding more completions and making the thing even more difficult to migrate).

Schottkyc137: One thing that's especially annoying is that all the test modules are in `vhdl_lang` and they can't be used in `vhdl_ls` (please correct me if I have overlooked a very simple way of doing this). Especially the `Code` object and related methods would be nice to use.

kraigher: Technically crates can have test-code in the public API as well. But the fact that the test utilities also needs to be ported to move the functionaly is a cause for concern. Previously I tried to put the UI/presentation layer in vhdl_ls and the pure language knowledge in the vhdl_lang crate. Maybe you could just move the CompletionItemKind etc to vhdl_ls? They feel out of place inside vhdl_lang crate. vhdl_lang really only needs to give a list of `AnyEnt` that can be completed and the mapping from `AnyEnt` to CompletionItem type could happen in vhdl_ls.

kraigher: PS: So to summarize in the current architecture the vhdl_lang contains knowledge about the vhdl language and vhdl_ls uses it to create a language server. Thus I would avoid having copies of lsp_types in vhdl_lang itself since it is an indication that this architecture is not followed.

Schottkyc137: I have done the refactoring. However, as you can see, the unit tests don't work because the `LibraryBuilder` and `Code` is not available outside of the `vhdl_lang` crate. What approach do you suggest for the `vhdl_ls` server? I can't seem to find a way to export `#[cfg(test)]` modules from `vhdl_lang`. Is there anything I have missed?

kraigher: I think you misunderstood what I meant by my previous comment. My point is that all the logic should be in `vhdl_lang` and the only thing that `vhdl_ls` does is to convert from `vhdl_lang` types into lsp-structures.  For example the code that creates all the completion options would be in vhdl_lang and tested in vhdl_lang. The only code that would be in vhdl_ls would be the code that converts an `AnyEnt` to a `CompletionItem`. The code that remains in `vhdl_ls` would then not be so important to test. 

For example you have `completion_item_from_decl` function. The code that is important to test is the one that generates the input to this function. The output of this function is just a conversion of a `vhdl_lang` internal type. 

If you look at how DocumentSymbol is implemented you will see that `vhdl_lang` tests all the logic needed to produce the document symbol and `vhdl_ls` is only the code that will convert the `vhdl_lang` internal representation into an lsp-type.

Schottkyc137: Alright I think I got your idea now so you'll probably like the current version :D

vhdl_ls/src/vhdl_server.rs
```diff
@@ -260,6 +261,84 @@ impl VHDLServer {
         }
     }
 
+    fn entity_kind_to_completion_kind(&self, kind: &AnyEntKind) -> CompletionItemKind {
```
kraigher, line 264: Why is this a method instead of a free function?

vhdl_ls/src/vhdl_server.rs
```diff
@@ -260,6 +261,84 @@ impl VHDLServer {
...
+            },
+            AnyEntKind::LoopParameter(_) => CompletionItemKind::MODULE,
+            AnyEntKind::PhysicalLiteral(_) => CompletionItemKind::UNIT,
+            AnyEntKind::DeferredConstant(_) => CompletionItemKind::CONSTANT,
+            AnyEntKind::Library => CompletionItemKind::MODULE,
+            AnyEntKind::Design(_) => CompletionItemKind::MODULE,
+        }
+    }
+
+    fn entity_to_completion_item(&self, ent: EntRef) -> CompletionItem {
```
kraigher, line 296: Does this need to be a method instead of a function?

vhdl_lang/Cargo.toml
```diff
@@ -25,6 +25,7 @@ parking_lot = "0"
 dunce = "1"
 pinned_vec = "0"
 itertools = "0"
+serde = { version = "1.0.156", features = ["derive"] }
```
kraigher, line 28: It feels a bit overkill to add a huge crate like serde as a dependency to `vhdl_lang` just to be able to serialize EntityId. Maybe Add a `toRaw/fromRaw`  method and let serde stay a dependency only in `vhdl_ls` where it is actually needed.

vhdl_lang/src/analysis/completion.rs
kraigher: Yes I am happy with the new version. It follows the architecture I was trying to explain. Very nice!

PR #209, opened 2023-11-04
Analyze Subprogram Instantiation Declarations

Schottkyc137:
- [x] Check that referenced subprogram exists
- [x] Check that referenced subprogram is an uninstantiated subprogram (and not a regular subprogram)
- [x] Check mismatch function / procedure instantiation
- [x] Check that a function call does not denote an uninstantiated subprogram
- [x] Resolve overloaded uninstantiated subprograms
- [x] Check generic map aspect
- [x] Declare the instantiated subprogram in the current scope
- [x] Set reference to the uninstantiated subprogram


vhdl_lang/src/analysis/named_entity/overloaded.rs
```diff
@@ -52,13 +66,19 @@ pub struct Signature<'a> {
     /// Vector of InterfaceObject or InterfaceFile
     pub(crate) formals: FormalRegion<'a>,
     pub(crate) return_type: Option<TypeEnt<'a>>,
+    pub(crate) generic_map: Option<Region<'a>>,
 }
 
 impl<'a> Signature<'a> {
-    pub fn new(formals: FormalRegion<'a>, return_type: Option<TypeEnt<'a>>) -> Signature<'a> {
+    pub fn new(
```
kraigher, line 73: Adding generics to the signature is a very invasive change. Are you sure this is the right approach?

Schottkyc137: I wasn't sure what the best place to store the generics is. Putting it here felt like the right place though as the generics can be thought of as a part of the signature. I initially also tried putting it into the `Overloaded` enum (creating an `UninstSubprogram` and `UninstSubprogramDecl` variant) but that was really tedious and cumbersome compared to the current solution.

kraigher: Isnt an uninstantiated subprogram supposed to be handled very differently from other subprogams? Adding enum variants to the Overloaded enum then seems like a good way to ensure as all existing code that handles normal subprograms should not also run for uninstantiated subprograms without careful consideration.

Schottkyc137: I have changed the implementation slightly. Now, only a `SignatureKey` has a boolean value `uninstantiated` to differentiate between uninstantiated entities. I cannot remove it there at the moment as this signature key needs to be unambiguous for overloaded entities. Do you think this approach is better?

Schottkyc137:
Alright, I think this is ready for review. I have also added some ignored tests that show the scope of this PR. All in all, this PR should cover most use cases. The ignored tests are mostly some edge-cases (e.g., recursive uninstantiated subprogram calls, uninstantiated subprogram calls as resolution function, e.t.c.)

vhdl_lang/src/analysis/tests/subprogram_instance.rs
```diff
@@ -0,0 +1,521 @@
...
+is begin
+end foo;
+
+procedure proc is new foo [bit];
+procedure proc2 is new foo [bit, bit];
+    ",
+    );
+
+    let diagnostics = builder.analyze();
+    check_no_diagnostics(&diagnostics);
```
kraigher, line 106: Doesnt the instantiation need to provide the type T?

kraigher:
I made a first overview of the code. Looks good and well tested. I will do a deeper review tomorrow.

vhdl_lang/src/analysis/tests/subprogram_instance.rs
Schottkyc137: Fixed.
The "better" test where the parameters are generic types and one can distinguish by signature is yet another edge case I didn't address in this PR. I have added another ignored test though where the desired behavior is shown.

vhdl_lang/src/analysis/tests/subprogram_instance.rs
```diff
@@ -0,0 +1,548 @@
+use crate::analysis::tests::{check_no_diagnostics, LibraryBuilder};
```
kraigher, line 1: New file lacks copyright notice

kraigher:
I had another look and I think it is solid. The only concern I have is it is really correct to add the `SignatureCategory` directly to the `SignatureKey`. It seems it is only necessary to be able to store uninstantiated subprograms and normal subprograms in the `OverloadedName`. Many other uses of `SignatureKey` should probably not care about `SignatureCategory` but this is then not reflected in the type which offers up the possibility of this field being something other than `Normal`. Would it not be more isolated to add a `SubprogramKey ` which was a wrapper arround `SignatureKey`:

```rust
struct SubprogramKey {
  cat: SignatureCategory,
  key: SignatureKey,
}
```

We would use `SubprogramKey` in only in the places where we actually need to avoid conflict between uninstantiated and instantiated subprograms.

PR #215, opened 2023-11-08
Fix visitor is slow for large designs

### LINKED ISSUE #212: Autocompletion / Intellisense not working for function arguments starting with version 0.67.0 (VSCode)
While typing in VSCode, it normally suggests autocompletion. Starting with VHDL-LS 0.67.0 triggering this autocompletions for function arguments fails in a larger project I'm working at.  Outside of function arguments, the feature works as intented.

I activated tracing messages in the server and it seems, that getting a response in this case takes more than 3 seconds which leads to the autocompletion feature failing. With VHDL-LS 0.66.0 the same response from the server only took 1 ms.

Here you see the behavior with VHDL-LS 0.66.0. Suggestions for autocompletion pop up immediately.
vhdl_ls_0 66 0_no_bug

This is the behavior with VHDL-LS 0.67.0 (also with all following up to 0.71.0). After hitting CTRL+SPACE to trigger the autocompletion manually, it seems to time out and fails:
vhdl_ls_0 67 0_bug

Unfortanately this is not reproducible with a minimal example small project and I can't share my project. If I should test something else or can help otherwise, please let me know.

Schottkyc137:
Closes #212. The `walk`  methods now only operate on a single file, and not the entire AST.

The issue can actually be checked with the example project. However, one must trigger completions inside a `MapAspect`. With this solution, completing takes a negligible amount of time (< 5 ms) compared to the pre 0.7.2 release version (~700ms).

I noticed another issue here, namely that some completions are not shown for some entities. Mostly this happens when the entity is not in the same design file, but not always. I think this has something to do with references not being set correctly but I haven't investigated further. I will open another issue for this.


vhdl_lang/src/analysis/root.rs
```diff
@@ -655,14 +655,23 @@ impl DesignRoot {
...
+    /// `UnitId`.
+    pub fn walk(&self, unit: &UnitId, visitor: &mut impl Visitor) {
+        let unit = self.get_unit(unit).unwrap();
+        if let Some(unit_rguard) = unit.unit.get() {
+            walk(unit_rguard.data(), visitor, &unit.tokens);
+        }
+    }
+
+    /// Walks all units in a source file denoted by `source`.
+    pub fn walk_source(&self, source: &Source, visitor: &mut impl Visitor) {
```
kraigher, line 668: A future improvement would be to walk just the design unit which maps to the EntityId. I hope to improve and unify the connection between the AST and the AnyEnts in the future.

vhdl_lang/src/analysis/completion.rs
```diff
@@ -63,9 +63,13 @@ struct PortsOrGenericsExtractor {
 
 impl DesignRoot {
     fn extract_port_or_generic_names(&self, id: EntityId, kind: MapAspectKind) -> Vec<EntityId> {
-        let mut searcher = PortsOrGenericsExtractor::new(id, kind);
-        self.walk(&mut searcher);
-        searcher.items
+        if let Some(ref pos) = self.get_ent(id).decl_pos {
+            let mut searcher = PortsOrGenericsExtractor::new(id, kind);
```
kraigher, line 67: Now that I look closer at what the PortsOrGenericsExtractor does it could be a lot simpler. You do not need to go via the AST to find the ports or generics of an entity. 

The `AnyEntKind::Component` contains a `Region` which contains all the `InterfaceEnt` for ports generics.
https://github.com/VHDL-LS/rust_hdl/blob/master/vhdl_lang/src/analysis/named_entity.rs#L45

The `AnyEntKind::Design(Design::Entity)` contains a `Region` which is all the ports and generics:
https://github.com/VHDL-LS/rust_hdl/blob/master/vhdl_lang/src/analysis/named_entity/design.rs#L23

kraigher:
Looks OK. I Wrote a comment that I realized the PortsOrGenericsExtractor  is completely uncessary. You do not need to walk the AST to find the generics and ports for a Entity/Component if you have their EntityId.

PR #216, opened 2023-11-08
Simplify the completion generics and port extraction method

Schottkyc137:
As was suggested, simplify the extraction method by removing the visitor and replacing it with the `AnyEntKind::Component` and `AnyEntKind::Design` values.

Also seems to fix the issue where completion options aren't shown when the declaration is in a separate file.

### DIFF
--- a/vhdl_lang/src/analysis/completion.rs
+++ b/vhdl_lang/src/analysis/completion.rs
@@ -1,15 +1,14 @@
-use crate::analysis::region::{AsUnique, NamedEntities};
+use crate::analysis::region::{AsUnique, NamedEntities, Region};
 use crate::analysis::{DesignRoot, HasEntityId};
 use crate::ast::visitor::{Visitor, VisitorResult};
 use crate::ast::{
-    AnyDesignUnit, AnyPrimaryUnit, AnySecondaryUnit, ComponentDeclaration, Designator,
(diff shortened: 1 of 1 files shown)


kraigher:
Looks good. Maybe it is not a problem but iterating over the items in the region will not return them in the source order. Just whatever order the hashtable stored them. If you need them sorted by their position in the list you need to sort them on srcpos.

Have a look here how I is done in another place of the codebase:
https://github.com/VHDL-LS/rust_hdl/blob/master/vhdl_lang/src/analysis/region.rs#L572

Schottkyc137:
> Looks good. Maybe it is not a problem but iterating over the items in the region will not return them in the source order. Just whatever order the hashtable stored them. If you need them sorted by their position in the list you need to sort them on srcpos.
> 
> Have a look here how I is done in another place of the codebase: https://github.com/VHDL-LS/rust_hdl/blob/master/vhdl_lang/src/analysis/region.rs#L572

I think that, in general, LSP clients sort according to alphabetical order (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion). This is not ideal though as, at the moment, all operators are shown before identifiers when completing `use` imports. Generally, however, identifiers are more common in imports compared to operators. The same ordering applies to completions for map aspects; they are sorted by the client. Not sure if there is a way around this as I also think that this isn't always what we want.

PR #223, opened 2023-11-13
Extend TOML example

Xcodo:
Add examples for wildcards/glob patterns and `is_third_party` setting in TOML file.

I think it's good to aid discoverability for these features, especially as common libraries like UVVM have many unused signals.

I can make the examples more generic if you prefer, but this one may also help some people understand how to support Xilinx libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ lib1.files = [
   'tb_ent.vhd'
 ]
 
+# Wildcards are supported
+lib3.files = [
+  'test/*.vhd',
+  'src/*.vhd',
+  'src/*/*.vhd',
+]
+
 # Libraries can be marked as third-party to disable some analysis warnings
 UNISIM.files = [
   'C:\Xilinx\Vivado\2023.1\data\vhdl\src\unisims\unisim_VCOMP.vhd',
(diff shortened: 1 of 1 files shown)


kraigher:
I think Unisim is also a very good example as I am sure it is a stumbling block for many new users. Some users might incorrectly assume they cannot use VHDL LS with Xilinx primitives because they do not know that there is just a file they can map.

To be really helpful I think also the readme.md in the VHDL LS vscode extension repo should be updated. I think most users look there for the examples: https://marketplace.visualstudio.com/items?itemName=hbohlin.vhdl-ls

README.md
```diff
@@ -61,6 +61,19 @@ lib1.files = [
...
 ]
+
+# Wildcards are supported
+lib3.files = [
+  'test/*.vhd',
+  'src/*.vhd',
+  'src/*/*.vhd',
+]
+
+# Libraries can be marked as third-party to disable some analysis warnings
```
kraigher, line 72: Maybe add an example of the warning ", such as unused declarations"

Xcodo:
Added that text. I'll do a similar PR in the extension repo soon.

PR #224, opened 2023-11-18
Feat: Emit diagnostic when a declaration is not allowed in the region it resides

### LINKED ISSUE #199: Don't allow declarations where they are not allowed
The following items are not allowed inside a subprogram declaration and could be flagged:

- [ ] Signals
- [ ] Shared Variables
- [ ] Component declaration

Similar examples can be found for other declarative parts.

Schottkyc137:
Closes #199 

Now correctly emits errors when, for example, a subprogram body is declared inside a package

### DIFF
--- a/vhdl_lang/src/analysis/declarative.rs
+++ b/vhdl_lang/src/analysis/declarative.rs
@@ -9,30 +9,18 @@ use super::*;
 use crate::ast::*;
 use crate::data::*;
 use crate::named_entity::{Signature, *};
-use crate::{ast, HasTokenSpan};
+use crate::{ast, named_entity, HasTokenSpan};
 use analyze::*;
 use fnv::FnvHashMap;
 use std::collections::hash_map::Entry;
 
-/// Describes where a declaration was made
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub enum DeclarativeContext {
-    Block,
-    Configuration,
-    Entity,
-    PackageBody,
-    Package,
-    Process,
-    ProtectedTypeBody,
-    Subprogram,
-}
-
 impl Declaration {
-    pub fn is_allowed_in_context(&self, ctx: DeclarativeContext) -> bool {
+    pub fn is_allowed_in_context(&self, parent: &AnyEntKind) -> bool {
         use Declaration::*;
         use ObjectClass::*;
-        match ctx {
-            DeclarativeContext::Block => matches!(
+        match parent {
+            AnyEntKind::Design(Design::Architecture(..))
+            | AnyEntKind::Concurrent(Some(Concurrent::Block | Concurrent::Generate)) => matches!(
                 self,
                 Object(ObjectDeclaration {
                     class: Constant | Signal | SharedVariable,
@@ -49,10 +37,10 @@ impl Declaration {
                     | Package(_)
                     | Configuration(_)
             ),
-            DeclarativeContext::Configuration => {
+            AnyEntKind::Design(Design::Configuration) => {
                 matches!(self, Use(_) | Attribute(ast::Attribute::Specification(_)))
             }
-            DeclarativeContext::Entity => matches!(
+            AnyEntKind::Design(Design::Entity(..)) => matches!(
                 self,
                 Object(_)
                     | File(_)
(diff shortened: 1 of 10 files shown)


vhdl_lang/src/analysis/declarative.rs
```diff
@@ -6,21 +6,94 @@
...
+}
+
 impl<'a> AnalyzeContext<'a> {
     pub fn analyze_declarative_part(
         &self,
         scope: &Scope<'a>,
         parent: EntRef<'a>,
         declarations: &mut [Declaration],
         diagnostics: &mut dyn DiagnosticHandler,
+        context: DeclarativeContext,
```
kraigher, line 96: Suggestion: The `parent` argument to the function is kidn of redundant with the new DeclarativeContext you added. The parent will be the package, block or whatever is the owner of the declarative part. So this code could potentially be simplified.

kraigher:
I wrote a suggestion which I think can simplify the code and remove a redundancy. 

A question is also if you have tried this on the example project and seen that runs cleanly? I usually test run all my changes on the example project to ensure it handles real world code.

Schottkyc137:
> I wrote a suggestion which I think can simplify the code and remove a redundancy.
> 
> A question is also if you have tried this on the example project and seen that runs cleanly? I usually test run all my changes on the example project to ensure it handles real world code.

I just noticed that I read the reference wrong when it comes to shared variables and normal variables. I thought that the production `variable` meant that only variables were allowed while `shared_variable` means only shared variables are allowed. However, it seems like `variable` means variables _and_ shared variables while `shared_variable` means only shared variables (not regular ones) are allowed. Will change in a bit!

vhdl_lang/src/analysis/declarative.rs
Schottkyc137: This is not always true though. For example, a `Block` or a `Process` simply passes the `parent` to the declarative part https://github.com/VHDL-LS/rust_hdl/blob/master/vhdl_lang/src/analysis/concurrent.rs#L99

Schottkyc137: A block or process do not always define a label. Should they still be named entities?

kraigher: The code is a bit complicated for concurrent and sequential statements since labels need to be visible before they are defined. Thus there is a bit of separation between the code which creates the AnyEnt for the block and the code which analyses the declarative part of the block. But I still think the parent is correct in this scenario. Otherwise the document symbols (outline in vscode) would be wrong.

Schottkyc137: Can you point out to me where the `AnyEnt` for the block is created? I only see the parent... (I pasted the wring line; it should be 103, sorry)

kraigher: I think they become named entities with Designator::Anonymous which I added just for DocumentSymbols. I think the anonymous labels even comes from the LRM but I am not sure and cannot check it now.

kraigher: The AnyEnt for the block is created in the self.define_labels_for_concurrent_part for the surrounding block.

kraigher: You can view this as an optional improvement suggestion from me that you should look into a bit. If it turns out that it is not as straightforward as I imagined we can keep your current solution with the extra argument.

As long as the example project works it is ok to merge.

Schottkyc137: I'll do it. I just didn't see that block was also always the correct `EnyEnt`. Example project now parses fine

Schottkyc137:
This has some conflicts with #209 because this PR introduces new `AnyEntKind`s, so this should probably be merged first.

vhdl_lang/src/analysis/tests/declarations.rs
```diff
@@ -0,0 +1,43 @@
+use crate::analysis::tests::{check_diagnostics, LibraryBuilder};
```
kraigher, line 1: This new file is missing a copyright header.

kraigher:
Looks good now, it is just missing a copyright notice in the new file.

Issue #219, opened 2023-11-12
Public API for traversing the design hierarchy / AST


kraigher:
We need to create a stable public API to traverse the design hierarchy and the AST. I am hoping @Schottkyc137 and @skaupper can contribute their ideas and feedback here. My goal is to replace the `Search` trait and the `Visitor` trait with this implementation. First we need to gather some requirements than we can discuss the implementation.

# Performance requirements
- Must not perform heap allocation
# Public requirements
## Library
- Must be able to get library by name
- Must be able to iterate over all libraries
## Design units
- Must be able to iterate over all design units in a source file
- Must be able to get primary units from library by name
- Must be able to iterate over all primary units in a libary
- Must be able to iterate over secondary units of a primary unit
- Must be able to get context clause of a design unit
## Declarations & Statements
- Must be able to iterate over all declarations, sequential statements, concurrent statements in a design unit
- Must be able to iterate over all declarations and statements nested within other declarations and statements
## Navigation
- Get referenced named entity (`AnyEnt`) from source location
- Get referenced design unit/statement/declaration in AST from named entity (`AnyEnt`)
- Get all source locations referencing named entity (`AnyEnt`)
# Private requirements
- Must be able to clear all references in the entire AST

Schottkyc137:
Some preliminary thoughts from myself without getting too much into details:

- Tool Directives
   Currently, tool directives are tokenized, but ignored afterwards so that no false positives are detected. However, a public API should make some methods / traits available to hook into the tokenizing process so that third-parties can utilize `vhdl_lang` to perform tasks on tool directives. Potential use-cases include the `protect` tool directive and a potential `vhdl_lang` tool directive to mute listing errors

- Scope
  I believe that this discussion could benefit from first defining use-cases of `vhdl_lang`. In my opinion, the design of a public API will benefit from making this clear at first. I can think of the following use-cases which would be of interred for me but others will probably be interested in this as well:
  - Tool directives for third-parties (see above)
  - Code Formatters
  - Documentation extractor (Referring to @skaupper's work)
  - Linters (should linting not be a concern of a public API since it's already part of the private API?)
  - Testbench generation or other code generation from VHDL
  - Code inspection in general (See #203 for example)
  - Integration with other HDL's / Language servers of other HDL's (There was interest in the Chat for some Verilog / VHDL shared language server)
  - Long-term goal: Simulator / full-fledged compiler for VHDL

Are any of these out of scope? Did I miss anything? What do you think?

[...]

PR #227, opened 2023-11-18
Potential new visitor candidate

Schottkyc137:
This visitor tries to avoid heap allocation by directly pushing onto the stack and not populating a `vec` first.

I tried to measure performance but the visitor has a small impact so I'm not sure how reliable this information is. However the number I give here were quite consistent:

- Without a Visitor: ~460 ms
- With a Visitor (old implementation): ~525 ms
- With a Visitor (new implementation): ~490 ms

The change seems to have an effect but it's not really noticeable (only around 30-40 ms). The visitor implementation I used:
```rust
struct Test {}

impl Visitor for Test {
    fn visit_ident(&mut self, _node: &Ident, _ctx: &dyn TokenAccess) -> VisitorResult {
        std::hint::black_box(());
        Continue
    }
}
```

I iterated over each source file in `root` and ran the visitor on the source file.

This PR is more of an idea concerning #219. That being said, it's fully functional without any API changes (so could theoretically be merged).

### DIFF
--- a/vhdl_lang/src/ast/visitor.rs
+++ b/vhdl_lang/src/ast/visitor.rs
@@ -8,7 +8,6 @@
 use crate::ast::visitor::VisitorResult::{Continue, Skip, Stop};
 use crate::ast::*;
 use crate::syntax::TokenAccess;
-use itertools::Itertools;
 use std::ops::Deref;
 
 #[derive(PartialEq)]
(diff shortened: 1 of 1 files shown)


kraigher:
I would say its not worth the extra complexity for the small gain in performance.

Schottkyc137:
> I would say its not worth the extra complexity for the small gain in performance.

What do you mean by extra complexity? The code size is almost the same. This also avoids the heap allocation (at least when iterating over the children)

kraigher:
Well complexity is relative. In the simplest code I could imagine for iterating over something we would just use an iterator. But here we need to push to a stack which is an extra argument. Sure it works but it also leaves me thinking if there is not a better way.
I will be away for a few hours now but I will come back and write more about this in the issue.

PR #230, opened 2023-11-25
extend goto/implementation to find architectures of entities

kernmatthias:
Hi, I've extended the find_implementation function to also search for architectures which implement an entity.


kraigher:
Looks good, you just forgot to run cargo fmt

kernmatthias:
Thank you for your review!

PR #232, opened 2023-11-26
[Suggestion] Split declarative analysis into smaller submodules

Schottkyc137:
While working with subprograms I have found it hard to navigate through the > 2000 Lines of code `declarative.rs` file. I suggest splitting this into smaller modules.
In this PR, the declaration analysis is split into the two modules `subprogram.rs` and `types.rs`.


vhdl_lang/src/analysis/types.rs
```diff
@@ -0,0 +1,647 @@
...
+        diagnostics: &mut dyn DiagnosticHandler,
+    ) -> FatalResult {
+        if let Err(err) = self.resolve_subtype_indication(scope, subtype_indication, diagnostics) {
+            err.add_to(diagnostics)?;
+        }
+        Ok(())
+    }
+}
+
+pub(crate) fn find_full_type_definition<'a>(
```
kraigher, line 628: Suggestion: This function is tightly coupled with the ` analyze_declarative_part` function and should remain private within the `declarative.rs`

kraigher (changes requested):

PR #238, opened 2023-12-17
Replace selected name with name

### LINKED ISSUE #147: Subtype error gives proken pkg

Schottkyc137:
Closes #147 

This removes the `SelectedName` enum. Analysis now only considers "regular" names. When only selected names are allowed in the specification, the parser still only considers selected names, but returns a regular name.
This simplifies the analyzer and removes duplication.

This has no measurable performance impact (tested using `cargo bench`)


Schottkyc137:
Will try to remove all occurrences of `resolve_selected_name`, therefore converting to draft

PR #243, opened 2023-12-21
Publish better diagnostic for attributes

### LINKED ISSUE #242: Incorrect diagnostic for attributes
The following code snippet:
```vhdl
entity test is
    attribute some_attr : string;
    attribute some_attr of test : signal is "some value";
end entity test;
```
produces the incorrect diagnostic "Attribute specification must be in the immediate declarative part". Instead, it should be something like "some_attr must be of class entity"

Schottkyc137:
Closes #242 

@kraigher is this the desired behavior? See the test below that I had to alter. Was setting `myattr` to `signal` instead of `entity` intentional?

### DIFF
--- a/vhdl_lang/src/analysis/declarative.rs
+++ b/vhdl_lang/src/analysis/declarative.rs
@@ -695,6 +695,9 @@ impl<'a> AnalyzeContext<'a> {
                 }
             };
 
+            // Attributes affect the underlying entity and cannot be set directly on aliases
+            let ent = ent.as_actual();
+
             if Some(*entity_class) != get_entity_class(ent) {
                 diagnostics.push(Diagnostic::error(
                     designator,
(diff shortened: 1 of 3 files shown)


vhdl_lang/src/analysis/declarative.rs
```diff
@@ -695,6 +695,14 @@ impl<'a> AnalyzeContext<'a> {
...
+            if Some(*entity_class) != get_entity_class(ent) {
+                diagnostics.push(Diagnostic::error(
+                    designator,
+                    format!("{} is not of class {}", ent.describe(), entity_class),
+                ));
+                return Ok(());
+            }
+
             // Attributes affect the underlying entity and cannot be set directly on aliases
             let ent = ent.as_actual();
```
kraigher, line 707: Blocking: Moving the entity class check above this line makes it no longer see through aliases.

Issue #164, opened 2023-02-21
[Feature] Instance placement autocompletion


pidgeon777:
I know that autocompletion is planned, my proposal is to add instance autocompletion.

For example, by typing the name of an entity that you want to instantiate, an autocomplete suggestion appears for that entity and selecting it would let you instantiate it as a snippet with placeholders (for example with VSCode format `$1`, `$0`). Pressing `Tab` \ `Shift+Tab` would let you place the cursor at the points where a keyword is needed. Example:

![g3c8lPdpcx](https://user-images.githubusercontent.com/13739892/220411531-6710c683-606c-4ec0-9514-9e57564aa7ba.gif)

SethGower:
@pidgeon777 I don't know what editor you use (it kinda looks like (neo)vim. If so, you could checkout this extension as a temporary solution to this: https://www.vim.org/scripts/script.php?script_id=3335

It lets you copy entity declarations and "paste" them as components/instances. I find it very helpful in my day to day, and have been using it for years. It's not the _most_ polished, but it does the job (mostly. Has some bugs though).

std-rubic:
A great feature, it seems to work with the 
https://github.com/VHDL-LS/rust_hdl/pull/244 
Thanks a lot !!

pidgeon777:
I have two proposal:

1. 

As you're aware, when defining generics and ports for a given entity in VHDL, they are arranged in an arbitrary order. For instance, the sequence might be `generic1`, `generic2`, `port1`, `port2`, `port3`, etc.

I propose that when entities are instantiated, the instantiated generics and ports should follow the same order as they were defined in their respective entities. So, autocomplete should output something as follows:

```vhdl
ENTITY_NAME: entity work.ENTITY_NAME
    generic map(
        generic1 => generic1,
        generic2 => generic2
    )
    port map(
        port1 => port1,
        port2 => port2,
        port3 => port3
    );
```

In the current implementation, it seems that the order is not preserved.

2.

An entity may have multiple architectures. In this case, the autocomplete should provide a list of architectures to choose from, put inside parentheses. For example:

```vhdl
ENTITY_NAME: entity work.ENTITY_NAME(ARCHITECTURE_1|ARCHITECTURE_2|ARCHITECTURE_3)
    generic map(
        generic1 => generic1,
        generic2 => generic2
    )
    port map(
        port1 => port1,
        port2 => port2,
        port3 => port3
    );
```

[...]

PR #244, opened 2023-12-25
Entity Completion

Schottkyc137:
Towards #164 

Unsure whether this closes 164, but this is a big steps towards it. This PR implements entity completion. What's still missing is

- showing components for instantiation
- ~enabling completing components and entities from other libraries (currently, everything is completed using the `work` library)~
- ~Limiting the scope. Currently, all visible entities are brought into scope~

There are also some other changes here (like adding the `TokenSpan` attributes to some statements) that I implemented but no longer need. However, these changes are beneficial anyway so I kept them. Hope this doesn't blow up the PR too much.

In theory, this PR is ready for review but I'll still keep it as draft since I want to add a couple of tests and comments.
Happy to receive feedback from anyone concerning the concrete form of completion.

### DIFF
--- a/vhdl_lang/src/analysis/concurrent.rs
+++ b/vhdl_lang/src/analysis/concurrent.rs
@@ -110,6 +110,7 @@ impl<'a> AnalyzeContext<'a> {
                     decl,
                     statements,
                     end_label_pos: _,
+                    ..
                 } = process;
                 if let Some(sensitivity_list) = sensitivity_list {
                     match sensitivity_list {
@@ -130,6 +131,7 @@ impl<'a> AnalyzeContext<'a> {
                     discrete_range,
                     body,
                     end_label_pos: _,
+                    ..
                 } = gen;
                 let typ = as_fatal(self.drange_type(scope, discrete_range, diagnostics))?;
                 let nested = scope.nested();
@@ -163,6 +165,7 @@ impl<'a> AnalyzeContext<'a> {
                             ref mut alternatives,
                         },
                     end_label_pos: _,
+                    ..
                 } = gen;
 
                 let ctyp = as_fatal(self.expr_unambiguous_type(scope, expression, diagnostics))?;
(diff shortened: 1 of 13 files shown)


kraigher:
I am away so I will review this some time after newyear.

Schottkyc137:
I'm going to mark this as ready for review for the time being. Component instantiation is something I still want to work on but I believe that this requires some more work. At least I didn't find a simple way to get all components for an architecture but maybe I'm just not using the language frontend correctly.
Component Instantiation is already supported by the language server; however there will be no components shown when completing.

pidgeon777:
If I understand correctly, this new behaviour will be as follows?:

https://github.com/VHDL-LS/rust_hdl/issues/164#issue-1593894310

Schottkyc137:
> If I understand correctly, this new behaviour will be as follows?:
> 
> [#164 (comment)](https://github.com/VHDL-LS/rust_hdl/issues/164#issue-1593894310)

Yes (at least that's the intended behaviour). The reason why #164 is not closed is because component instantiation is still missing. I hope to implement that and more completions soon.

pidgeon777:
As you're aware, when defining generics and ports for a given entity in VHDL, they are arranged in an arbitrary order. For instance, the sequence might be generic1, generic2, port1, port3, port2, etc.

I propose that when entities are instantiated, the instantiated generics and ports should follow the same order as they were defined in their respective entities. Do you think this is feasible? In the current implementation, it appears that the order differs.

Schottkyc137:
> As you're aware, when defining generics and ports for a given entity in VHDL, they are arranged in an arbitrary order. For instance, the sequence might be generic1, generic2, port1, port3, port2, etc.
> 
> I propose that when entities are instantiated, the instantiated generics and ports should follow the same order as they were defined in their respective entities. Do you think this is feasible? In the current implementation, it appears that the order differs.

I think this should definitely be the case. However, that behaviour should actually already have been implemented. Unfortunately, I cannot reproduce the arbitrary order. What I see is that ports are omitted that contain errors. For example, if a type name of a port or generic could not be resolved, the port will not appear in the instantiated entity. Do you have an example where the random order appears?

Issue #246, opened 2024-01-03
No declaration within package instance


ncervino:
Hello,

The code sample below causes VHDL-LS to generate a problem report with a message of `No declaration of 'pkg1' within package instance 'pkg2'`.   The sample compiles without errors or warnings using ModelSim.  Thanks for your help!

``` VHDL
library ieee;
use ieee.std_logic_1164.all;

package pkg1 is
    generic(i_gen: natural);

    type type1_t is array (0 to i_gen) of std_logic_vector(i_gen - 1 downto 0);
end package pkg1;

package pkg2 is
    generic (i_gen : natural);

    package pkg1 is new work.pkg1 generic map (i_gen);

    type type2_t is array (0 to i_gen) of pkg1.type1_t;
end package pkg2;

entity ent is
end entity ent;

architecture arch of ent is
    package pkg2 is new work.pkg2 generic map (1);
begin
    process is
        variable v2 : pkg2.type2_t;
        variable v1 : pkg2.pkg1.type1_t;
    begin
        wait;
    end process;
end architecture arch;
```

PR #247, opened 2024-01-07
Distinguish between packages that are instantiated in a generic map and in the declarative region

Schottkyc137:
Towards #246

In the current implementation, Package instances are always assumed to be in the generic map of another package, i.e.
```vhdl
package foo is
    generic (
        package x is new bar generic map (<>)
    );
end package;
```
However, this is cannot always be assumed as the following statement is also valid:
```vhdl
package foo is
    package x is new work.pkg2 generic map (0);
end package;
```

This PR distinguishes these two cases by adding a boolean parameter to `Design::PackageInstance`.
The only difference at the moment is that the error message from #246 is no longer `No declaration of 'foo' within package instance 'bar'`, but `package instance cannot be selected from design unit`. As a consequence, I haven't written any tests as the only consequence of this PR can be seen in the error message which is incorrect. Tests will be added to another PR.

### DIFF
--- a/vhdl_lang/src/named_entity/design.rs
+++ b/vhdl_lang/src/named_entity/design.rs
@@ -22,7 +22,13 @@ pub enum Design<'a> {
     Package(Visibility<'a>, Region<'a>),
     PackageBody,
     UninstPackage(Visibility<'a>, Region<'a>),
-    PackageInstance(Region<'a>),
+    // The second parameter of `PackageInstance` is `true`, when this package instance is
+    // part of a generic map, i.e.,
+    // generic map (
+    //     package foo is new bar generic map (<>)
+    // );
+    // otherwise, the instantiated packages appears regularly in the declarative region.
+    PackageInstance(Region<'a>, bool),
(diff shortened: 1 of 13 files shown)


vhdl_lang/src/named_entity/design.rs
```diff
@@ -22,7 +22,13 @@ pub enum Design<'a> {
...
     PackageBody,
     UninstPackage(Visibility<'a>, Region<'a>),
-    PackageInstance(Region<'a>),
+    // The second parameter of `PackageInstance` is `true`, when this package instance is
+    // part of a generic map, i.e.,
+    // generic map (
+    //     package foo is new bar generic map (<>)
+    // );
+    // otherwise, the instantiated packages appears regularly in the declarative region.
+    PackageInstance(Region<'a>, bool),
```
kraigher, line 31: I think it would be clearer to have a separate InterfacePackage or simial as a new enum variant instead of having an extra boolean.

kraigher (changes requested):

PR #260, opened 2024-02-11
Add aarch64-apple-darwin as target architecture

Schottkyc137:
Adds apple silicon based machines as target


Schottkyc137:
Note that this PR does not allow distributing the executable in VSCode as this needs [Notarization](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution). One can get around that but it's somewhat tricky and not a nice user experience.

PR #265, opened 2024-03-01
Substitute environment variables into config paths

### LINKED ISSUE #174: [Feature] Add support to environment variable usage in the vhdl_ls.toml
It would be helpful to be able to use environment variables in the `vhdl_ls.toml` file, so:

1. paths are not hardcoded
2. easier to share the same `vhdl_ls.toml` file across different engineers working on the same project
3. easier integration with CI
4. avoid effort duplication writing the `vhdl_ls.toml` 

expected result:

```toml
[libraries]
work.files = [
	'rtl/*.vhd',
	'sim/*.vhd',
]

vunit_lib.files = [
	'$VUNIT_PATH/vunit/vhdl/*.vhd',
	'$VUNIT_PATH/vunit/vhdl/com/src/*.vhd',
	'$VUNIT_PATH/vunit/vhdl/run/src/*.vhd',
	'$VUNIT_PATH/vunit/vhdl/core/src/*.vhd',
	'$VUNIT_PATH/vunit/vhdl/check/src/*.vhd',
	'$VUNIT_PATH/vunit/vhdl/path/src/*.vhd',
	'$VUNIT_PATH/vunit/vhdl/random/src/*.vhd',
	'$VUNIT_PATH/vunit/vhdl/logging/src/*.vhd',
	'$VUNIT_PATH/vunit/vhdl/data_types//src/*.vhd',
	'$VUNIT_PATH/vunit/vhdl/osvvm/*.vhd',
]
```

KarelPeeters:
Fixes #174.

I've tested this PR in the VSCode plugin using the Language Server User Path, and it works great for me.

### DIFF
--- a/vhdl_lang/src/config.rs
+++ b/vhdl_lang/src/config.rs
@@ -8,6 +8,7 @@
 
 use crate::data::*;
 use fnv::FnvHashMap;
+use std::collections::HashMap;
 use std::env;
 use std::fs::File;
 use std::io;
@@ -102,6 +103,8 @@ impl LibraryConfig {
 
 impl Config {
     pub fn from_str(string: &str, parent: &Path) -> Result<Config, String> {
+        let env_vars: HashMap<String, String> = std::env::vars().collect();
+
         let config = string.parse::<Value>().map_err(|err| err.to_string())?;
         let mut libraries = FnvHashMap::default();
 
@@ -130,6 +133,8 @@ impl Config {
                 let file = file
                     .as_str()
                     .ok_or_else(|| format!("not a string {file}"))?;
+
+                let file = substitute_environment_variables(file, &env_vars);
 
                 let path = parent.join(file);
                 let path = path
@@ -276,6 +281,34 @@ impl Config {
         self.load_home_config(messages);
         self.load_env_config("VHDL_LS_CONFIG", messages);
     }
+}
+
+fn substitute_environment_variables(s: &str, replace: &HashMap<String, String>) -> String {
+    let mut result = String::new();
+
+    let mut left = s;
+    while let Some(start) = left.find('$') {
+        // keep non-env var piece as-is
+        result.push_str(&left[..start]);
+        left = &left[start + 1..];
+
+        // replace env var
+        let env_name_len = left
+            .find(|c: char| !(c == '_' || c.is_ascii_alphanumeric()))
+            .unwrap_or(left.len());
+        let env_name = &left[..env_name_len];
+
+        let replacement = replace.get(env_name).map(String::as_ref).unwrap_or("");
+        result.push_str(replacement);
+
+        // skip past env var
+        left = &left[env_name_len..];
+    }
+
+    // keep remaining string
+    result.push_str(left);
+
+    result
 }
 
 /// Returns true if the pattern is a plain file name and not a glob pattern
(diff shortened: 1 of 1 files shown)


KarelPeeters:
I'm not sure what the best way to handle non-existing environment variables is. Right now I just assume a default value of `""` like bash without `set -u`, but maybe another option is better:
* explicitly show an error to the user
* silently ignore any paths that contain an undefined variable. This would also work as a convenient way to have optional libraries in the config file.

vhdl_lang/src/config.rs
```diff
@@ -102,6 +103,8 @@ impl LibraryConfig {
 
 impl Config {
     pub fn from_str(string: &str, parent: &Path) -> Result<Config, String> {
+        let env_vars: HashMap<String, String> = std::env::vars().collect();
```
Schottkyc137, line 106: This creates a copy of all environment variables. I assume that this is in order to test the code? I would prefer it if this wasn't the case, but if it's too much of a hassle you can leave it as-is

vhdl_lang/src/config.rs
```diff
@@ -278,6 +283,34 @@ impl Config {
...
+    let mut left = s;
+    while let Some(start) = left.find('$') {
+        // keep non-env var piece as-is
+        result.push_str(&left[..start]);
+        left = &left[start + 1..];
+
+        // replace env var
+        let env_name_len = left
+            .find(|c: char| !(c == '_' || c.is_ascii_alphanumeric()))
+            .unwrap_or(left.len());
```
Schottkyc137, line 298: If the string wasn't found, for example because the env is something illegal like `$0abc`, I would prefer this to be communicated to the user.

vhdl_lang/src/config.rs
```diff
@@ -278,6 +283,34 @@ impl Config {
...
+        let env_name = &left[..env_name_len];
+
+        let replacement = replace.get(env_name).map(String::as_ref).unwrap_or("");
```
Schottkyc137, line 301: Same as above, I would prefer if this is communicated to the user.

Schottkyc137 (commented):
Thanks a lot for your contribution! This is a feature that many users will like.

I have added a few comments. None of them are blocking, so you may choose to address them but you can also ignore them.

vhdl_lang/src/config.rs
KarelPeeters: That's true, but that's just how `std::env::vars()` works: it returns an iterator over `String` pairs. There's no way to avoid the copy that implies.

The alternatives are:
* Do this only once at plugin startup and store it in a `lazy_static` or similar, but that's annoying.
* Call `std::env::vars()` for each substitution separately, but that makes even more copies.

vhdl_lang/src/config.rs
KarelPeeters: I overlooked that error case, I'll fix it.

vhdl_lang/src/config.rs
KarelPeeters: I agree, I'll add a warning.

vhdl_lang/src/config.rs
Schottkyc137: Did you consider [std::env::var](https://doc.rust-lang.org/std/env/fn.var.html) for this use-case?

KarelPeeters: I forgot about that one! To keep testing working maybe we can take an `Fn(K) -> Result<String, VarError>` as a parameter and then pass `std::env::var` for the real use case?

Schottkyc137: That sounds like a nice solution!

vhdl_lang/src/config.rs
KarelPeeters: There's a mix of two error handling mechanisms used in this file:
* `Config::from_str` can report a single error through the `Result` it returns, which also marks a total failure to load the file.
* Most other functions report errors or warnings through `MessageHandler` and keep going with a best effort.
Which option should be used for these env var errors? I think the latter fits better but that requires adding a `messages` parameter to `Config::from_str`, is that okay?

Schottkyc137: Given that all other issues with the `vhdl_ls.toml` file also cause an error I would prefer the first option you proposed.

KarelPeeters:
I think I've addressed everything.

Xcodo:
Thanks for doing this PR 👍. This was a problem for my team this week, fantastic to see a fix I can roll out soon!

KarelPeeters:
Thanks for the quick feedback and merge!

Issue #266, opened 2024-03-01
In the readme.md add a description how to install VHDL_LS for Neovim


Chris44442:
In the readme.md add a description how to install VHDL_LS for Neovim:

## Use in Neovim

Neovim provides an LSP client to the VHDL_LS language server. Install the vhdl_ls  
binary e.g. via Cargo. The binary must be on the path and executable (if you  
can run "vhdl_ls -h" in the terminal then you're good). Put the "vhdl_libraries"  
directory into the parent directory of the vhdl_ls binary. 

In your Neovim config.lua add the following:  

```lua
function STARTVHDLLS()
  vim.lsp.start({
    name = 'vhdl_ls',
    cmd = {'vhdl_ls'},
  })
end
vim.api.nvim_set_keymap('n', '<F5>', ':lua STARTVHDLLS()<CR>', { noremap = true, silent = true })
```

This is just one example, there are many ways to run language servers. Anyway  
when you are inside a buffer you can now hit F5 to start VHDL_LS. Autostart is  
also possible. For more info, have a look at the [Neovim LSP documentation](https://neovim.io/doc/user/lsp.html).

Schottkyc137:
This sounds like a great addition! Can you open a Pull Request with the suggested change?

PR #269, opened 2024-03-03
added "Use in Neovim" to readme

Chris44442:
[Issue#266](https://github.com/VHDL-LS/rust_hdl/issues/266#issue-2164238029)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,12 +52,11 @@ following snippet to your config:
 (vhdl-ext-lsp-set-server 've-rust-hdl)   ; `lsp' config
 ```
 
-## Use in Neovim
+## Manual Installation for Neovim
 
-Neovim provides an LSP client to the VHDL_LS language server. Install the VHDL_LS  
-binary e.g. via Cargo. The binary must be on the path and executable (if you  
-can run "vhdl_ls -h" in the terminal then you're good). Put the "vhdl_libraries"  
-directory into the parent directory of the VHDL_LS binary. 
+Neovim provides an LSP client to the VHDL_LS language server. Download the  
+VHDL_LS release. The binary must be on the path and executable (if you can run  
+"vhdl_ls -h" in the terminal then you're good).
 
 In your Neovim config.lua add the following:  
 
@@ -71,9 +70,9 @@ end
 vim.api.nvim_set_keymap('n', '<F5>', ':lua STARTVHDLLS()<CR>', { noremap = true, silent = true })
 ```
 
-This is just one example, there are many ways to run the language server. Anyway  
-when you are inside a buffer you can now hit F5 to start VHDL_LS. For more info  
-and options, have a look at the [Neovim LSP documentation](https://neovim.io/doc/user/lsp.html).
+Using the example above, pressing F5 while inside Neovim starts the language  
+server. There are also other options, like automatically starting it when  
+opening a certain file type, see the [Neovim LSP documentation](https://neovim.io/doc/user/lsp.html) for more.
 
 ### Configuration
 The language server needs to know your library mapping to perform full analysis of the code. For this it uses a configuration file in the [TOML](https://github.com/toml-lang/toml) format named `vhdl_ls.toml`.
--- a/README.md
+++ b/README.md
@@ -52,6 +52,29 @@ following snippet to your config:
 (vhdl-ext-lsp-set-server 've-rust-hdl)   ; `lsp' config
 ```
 
+## Use in Neovim
+
+Neovim provides an LSP client to the VHDL_LS language server. Install the VHDL_LS  
+binary e.g. via Cargo. The binary must be on the path and executable (if you  
+can run "vhdl_ls -h" in the terminal then you're good). Put the "vhdl_libraries"  
+directory into the parent directory of the VHDL_LS binary. 
+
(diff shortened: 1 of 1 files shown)


Schottkyc137:
Thanks for the contribution!

At the moment, the instructions sound just a little bit too generic to me. Can you change the tone so that it is clear that this is only one (of possibly many) options to enable the language server? For example, there is a [Pull Request](https://github.com/mason-org/mason-registry/pull/4596) trying to add `vhdl_ls` to Mason.
Also, instead of using `cargo`, one could compile from source or download the pre-built binaries.

README.md
```diff
@@ -52,6 +52,29 @@ following snippet to your config:
 (vhdl-ext-lsp-set-server 've-rust-hdl)   ; `lsp' config
 ```
 
+## Use in Neovim
+
+Neovim provides an LSP client to the VHDL_LS language server. Install the VHDL_LS  
+binary e.g. via Cargo. The binary must be on the path and executable (if you  
+can run "vhdl_ls -h" in the terminal then you're good). Put the "vhdl_libraries"
```
Schottkyc137, line 59: I'm not quite sure everyone understands what The "vhdl_libraries" are. Could you add a small explanation here?

README.md
```diff
@@ -52,6 +52,29 @@ following snippet to your config:
...
+function STARTVHDLLS()
+  vim.lsp.start({
+    name = 'vhdl_ls',
+    cmd = {'vhdl_ls'},
+  })
+end
+vim.api.nvim_set_keymap('n', '<F5>', ':lua STARTVHDLLS()<CR>', { noremap = true, silent = true })
+```
+
+This is just one example, there are many ways to run the language server. Anyway
```
Schottkyc137, line 74: What are the other ways to run the language server?

Schottkyc137 (changes requested):
To elaborate more on my comment: I think this is more or less one example of how you could integrate `vhdl_ls` into neovim, correct? If so I'd appreciate it if the text would reflect this.

Chris44442:
Oh, I didn't know about the [Mason PR](https://github.com/mason-org/mason-registry/pull/4596). Yeah, this would be the preferred way to do it of course. But until its merged, I guess it makes sense to describe a way, how to manually add the LS to Neovim. I know it would've saved me some time a few months back when I started using it and didn't know what I was doing.

> I think this is more or less one example of how you could integrate vhdl_ls into neovim, correct?

At the end of the day they all use the vim.lsp framework. But I will try to make it clear that there is an automatic way upcoming (mason), and that this here is the way to do it manually. If that is ok with you.

README.md
Chris44442: What I meant is that you can also autorun the LS, or start it when you enter a buffer, or tie it to another plugin somehow. Essentially any way you like within what Neovim allows. But wanted to keep the "How to" short. Should I just remove this remark?

README.md
Chris44442: Thanks for bringing that up, it is something I didn't quite fully understand. There is the [vhdl_libraries](https://github.com/VHDL-LS/rust_hdl/tree/master/vhdl_libraries)  directory in this repo. When I use VHDL_LS the way I described it expects this directory in the parent directory of the VHDL_LS binary, otherwise it will crash in Neovim. Simply copying it there worked for me, but I guess it is not exactly how it is supposed to be, right?

Schottkyc137:
> At the end of the day they all use the vim.lsp framework. But I will try to make it clear that there is an automatic way upcoming (mason), and that this here is the way to do it manually. If that is ok with you.

Yea, that sounds good. My only concern is that the, way you described it, it sounds a bit like that's the only way of installing the client for nvim. I'm already happy with changing the title to `Manual Installation for NeoVim` (or something in that direction) to signify that there are other methods.

README.md
Schottkyc137: I like it, but maybe rephrase it to something like "Using the example above, pressing F5 while inside NeoVim starts the language server. There are also other options, like automatically starting it when opening a certain file type, see the doc for more..."

README.md
Schottkyc137: Well, yes and no :D
The language server needs the standard libraries to correctly work. That way it can, for example, resolve `std_logic` or other symbols from the standard library when you use them. So therefore simply copying is fine. However, the better approach would be to package the libraries with the executable. This is done for the artefacts and when you download a release, you get the packaged libraries with that release. However, using `cargo`, there is no standard way of adding additional files to the binary release and package them. Therefore, this is a bit awkward at the moment.

Chris44442: ohhh I see. Thanks for clearing that up. Maybe this should be noted somewhere else in the readme then? Because I am probably not the only one who installed it via cargo.

Schottkyc137:
I'm happy with the changes now. This is a great addition for anyone trying to use Neovim with `vhdl_ls`. If you're happy too I will merge this.

README.md
Schottkyc137: I agree. This should probably be noted down somewhere (preferably the readme) and the error message should point to the readme when the libraries cannot be found,

Chris44442:
> If you're happy too I will merge this.

Yes, all good. Thanks!

Issue #278, opened 2024-03-18
[BUG] File paths with $ are incorrectly being interpreted as environment variables


Matt-D-Smith:
If a file path has a '$' character in it, the parser for vhdl_ls.toml interprets it as an environment variable when it should not. $ characters are valid in windows file path names, and used in network/shared file paths to indicate hidden folders.

Example entry in vhdl_ls.toml
```
unisim.files = [ 
   '\\networklocation\cad$\apps\xilinx_vitis\Vivado_2020.2\Vivado\2020.2\data\vhdl\src\unisims\unisim_VCOMP.vhd',
]
```

Resulting error in VHDL LS output:
```
Error loading vhdl_ls.toml: environment variable '' is not defined
```

Schottkyc137:
This is fixed with the latest commit by disabling the substitution feature for windows machines. However the current solution is unsatisfactory as windows users would likely also want to have this feature, probably using the `%VAR_NAME%` syntax instead of the `$VAR_NAME` syntax. I have opened [an issue](https://github.com/fizyr/subst/issues/17) on the `subst` package addressing this.

Xcodo:
Yes I'm quite disappointed by that change - the variable substitution
feature was fantastic for us and we're on Windows.

Perhaps an alternative would be to have a setting to enable or disable the
substitution?

PR #281, opened 2024-03-23
Implement variable substitution for Windows machines

Schottkyc137:
See #278

Now also allows variable substitution in Windows using the `%VAR_NAME%` syntax

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,8 @@ UNISIM.is_third_party = true
 ```
 
 Paths in the `vhdl_ls.toml` can contain glob patterns (i.e., `.../*/`).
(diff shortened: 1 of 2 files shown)


Schottkyc137:
@Xcodo do you think that this is would work for Windows users?

Xcodo:
Looks reasonable to me, thanks for implementing that.

I had a look at the conversation in the [subst package](https://github.com/fizyr/subst/issues/17) too. I don't think `%VAR_NAME%` is absolutely required. `%{VAR_NAME}` would also likely be fine for most people as long as it is documented.

Schottkyc137:
I'll merge this for now just to be able to roll the fix for windows out. But this will likely not a very long-lived change, depending on what `subst` does.

Issue #255, opened 2024-01-30
[FEATURE] Add exclusion list for problems panel


nohrel:
**Feature Description** :

Let user add a list of files for which no errors/warnings should be raised.

**Feature Usecase** :

Defining libraries in "vhdl_ls.toml" which raise tons of errors due to false errors seen by VHDL_LS.
- Those libraries need to be parsed for project, but they can be ignored for syntax errors 

Examples of errors that can currently be triggered : 
- Functions overriding with "std_logic_vector" and "std_ulogic_vector" -> "Duplicate declaration of "FunctionName" with signature "std_logic_vector"
_Both types are detected as identical

- Encrypted library files -> "Wrong Token" 

- etc...

befedo:
Hi @nohrel 

Isn't `is_third_party` the switch you're looking for?

from the Readme:
> Libraries can be marked as third-party to disable some analysis warnings, such as unused declarations

nohrel:
Hi @befedo 

Unfortunately, the `is_third_party` switch does not solve this issue :

- It still shows "third-party" libraries errors (seems to only disable "some warnings")
- I would also like to "exclude" a subset of files from the "work" library (without disabling analysis for all files in work library)

Examples of errors raised using the `is_third_party` switch, on ModelSim libraries (modelsim_lib, std_developerskit)

![image](https://github.com/VHDL-LS/rust_hdl/assets/158173163/9994b43b-854e-4b9e-b2ef-38696a47403e)

kraigher:
is_third_party only disables the unused code detection, not all errors.

You almost never want unused code warnings from third party code. Errors however you need even for third party code. The third party code still needs to be understood by the tool to check your use of it.

nohrel:
My understanding is that, when VHDL_LS identifies errors in a `.vhd` file : 
- When the file is a "work library" module (entity/arch), VHDL_LS will raise errors in **all** other `.vhd` files instantiating that module 
- When the file is a "work library" package, or when the file is from another library, VHDL_LS will not raise errors for other `.vhd` files using the package or library

It seems that "excluding" some `work` packages, or `third party libraries`, from the Error Report Panel would not have an impact on VHDL_LS capability to correctly parse the main code that uses them 
(example : constants, functions, procedures, components are still recognized in files with errors).

Regarding the external tool chain (simulation, synthesis), it seems that few of their own VHDL libraries include some VHDL syntax errors seen by VHDL_LS without impacting their tool. 

[...]

Issue #192, opened 2023-08-22
[Feature] Add option to mute the error and/or disable lint checking


nselvara:
I'd appreciate a feature that allows to mute a warning/error and/or disable lint checking/code analysis for particular lines. Maybe to mute particular lines, we could add for example some kind of comments in the code.

```vhdl
-- lint_off
variable some_var: integer;
-- lint_on
```

Schottkyc137:
At the moment I don't think that this will be implemented. `vhdl_ls` is not a linter. If something is flagged as wrong, this means that it doesn't comply with the standard. Do you have a concrete use-case in mind?

nselvara:
Sorry, I didn't mean in particular the linter but the static code analysis. As an example this one could serve: https://github.com/VHDL-LS/rust_hdl/issues/184
Or this one: https://github.com/VHDL-LS/rust_hdl/issues/191

nselvara:
Sorry, I didn't mean in particular the linter but the static code analysis. As an example this one could serve: https://github.com/VHDL-LS/rust_hdl/issues/184
Or this one: https://github.com/VHDL-LS/rust_hdl/issues/191

Schottkyc137:
Ah, I see. Yes, this would make more sense. However, I think that implementing this feature ist as hard or even harder than fixing the mentioned feature as there is currently no comprehensive way to get the tokens (and by that token [pun intended] no comprehensive way to get comments and/or tool directives for a certain line). An easier approach could be to turn on/off analysis for a file or a for a project completely. Would that be a solution?

nselvara:
I see, yeah I mean for the mean time that would be a sufficient solution. It's more like for developers who work on those projects could be a hint that the line is actually correct but the plugin flags it as an error. I'm not sure if that's maybe easier but what instead writing a comment but rather selecting lines of code and exculding those from the analysis. I think if the file/project can be excluded there must be some kind of flags that are stored so I assume that some lines of numbers can be somehow also included? I don't know hard this solution might be to implement.

darsor:
This feature would be very useful for PSL code, which is not currently supported by VHDL-LS.

PR #282, opened 2024-04-01
Add Diagnostic Error Codes

Schottkyc137:
# Motivation

There are a lot of issues about the general concept of allowing more message control (i.e., #255, #192, and some in the [rust_hdl_vscode](https://github.com/VHDL-LS/rust_hdl_vscode/issues/16) repository). This is an attempt at tackling the challenge of allowing message control of sorts. Once all is set and done, I envision a process where you can
1) Disable or downgrade diagnostics in the `vhdl_ls.toml`, i.e., 
  ```toml
  [libraries]
  [lib1.files] = [...]

  [config]
  [error_codes]
  # Downgrade all errors of the following kind to warnings
  misplaced_attribute_spec = "warning"
   ```
2) Disable or downgrade diagnostics in vhdl using the tool directive concept or plain comments:
  ```vhdl
  entity foo is
      port (
          `vhdl_ls allow=unused
          bar: in bit;
          -- vhdl_ls allow=unused
          baz: out bit
      );
  end entity;
  ```

(Note that the syntax for both examples is just for illustrative purposes and subject to change should these features be implemented).

# This PR
This PR introduces error codes. Except for being forwarded to the language server and potentially being displayed, this is unused. However, I plan to add additional functionality that uses these error codes. This is to avoid error prone regex matching or similar when detecting an error that should be ignored or downgraded:

```vhdl
entity foo is
    port (
         -- Might work for now but matching on the text is potentially bad for readability and the actual message being displayed is subject to change
        `vhdl_ls disable="unused declaration pf port '*'"
        bar: in bit
    );
end entity;
```

I will leave this PR open for some time to collect some feedback, especially concerning how current errors without error codes are assigned to the different error codes. Later changes to the actual codes that are not purely additive (i.e., merging of two error codes, splitting an error code into multiple, ...) are not desirable as this will cause backwards incompatible changes. If no such feedback is given, I will assume that these error codes are a good start.


kraigher:
I think it is a good step to create error codes if we want to support user configurable error ignores.
I think it is very important that the error code is referred to by name and not by some magic number by the user. Just like in Rust with [allow(error_code)].  You could generate the error names by using the variant names of the rust enum, that will be very easy to maintain.

Schottkyc137:
Yea, from and to string are definitely planned. I'm not sure whether to use a custom string - to enum (and inverse) map or auto-deriving is better. The first approach has the advantage that the code representation of the enum is independent of what is seen by the user. One could, for example, easily change the name of the enum while at the same time being backwards compatible. But then on the other hand for packages like [strum](https://crates.io/crates/strum_macros), such features are already present, I believe (by adding aliases for names). So I'm also heavily leaning towards auto generating names.

kraigher:
Auto generating names is probably the most maintainable. With strum you can have aliases for exceptions.

Schottkyc137:
Probably there is little benefit now from having the numeric value being displayed. Currently, for example in VSCode, the error is displayed as `"No declaration of ..." vhdl_ls(E038)`. But it's probably better to display it as `"No declaration of ..." vhdl_ls(undeclared)`. Or do you see any benefits in keeping the numeric values, @kraigher?

kraigher:
Good systems for error codes do not use numbers. They add no value. A short name is both unique, self describing and can be looked up in a reference doc.

PR #283, opened 2024-04-01
Enable configurable severities

### LINKED ISSUE #162: [Enhancement] Add more settings to `rust_vhdl` project file
[VHDL-Tool](https://www.vhdltool.com/) is currently making use of a `vhdltool-config.yaml` project file as follows:

```yaml
#Define your project's libraries and source files here.
#This section is compulsory.
Libraries:
    #The name of the library.
    - name: hardware_lib
    #The paths where the source files for this library can be found. Use "**" to match arbitrarily nested directories.
      paths: 
        - "hardware_lib/**/*.vhd"
        - "hardware_lib/**/*.vhdl"


    - name: hardware_lib2
      paths: 
        - "hardware_lib2/**/*.vhd"
        - "hardware_lib2/**/*.vhdl"


    #Point to the IEEE standard libraries
    - name: ieee
      paths:
        - "/home/awesome-vhdl-dev/ieee/*.vhd"
        - "/home/awesome-vhdl-dev/ieee/*.vhdl"


Preferences:
    #Enable/disable typechecking
    TypeCheck: True


    #Display errors over multiple lines
    MultiLineErrors: True


    #Enable/disable check-as-you-type
    CheckOnChange: True


    #Enable/disable linting
    Lint: True


    #Display the first sintax error only in a file
    FirstSyntaxErrorOnly: True


#Linter rule configuration.
#Rules can be enabled or disabled.
#Rules also have a severity. It may be one of Info, Warning, Critical or Error.
Lint:
    #Threshold, below which messages are not displayed.
    Threshold: Warning


    #Long form rule configuration. Both enabled/disabled status and severity can be configured this way.
    DeclaredNotAssigned: 
        enabled:  True
        severity: Warning #Default severity Warning


    #Short form. Only enabled/disabled status can be specified. Severity is the default for the rule.
    DeclaredNotRead:             True #Default severity Warning
    ReadNotAssigned:             True #Default severity Critical
    SensitivityListCheck:        True #Default severity Warning
    ExtraSensitivityListCheck:   True #Default severity Warning
    DuplicateSensitivity:        True #Default severity Warning
    LatchCheck:                  True #Default severity Critical
    VariableNotRead:             True #Default severity Warning
    VariableNotWritten:          True #Default severity Warning
    PortNotRead:                 True #Default severity Warning
    PortNotWritten:              True #Default severity Critical
    NoPrimaryUnit:               True #Default severity Warning
    DuplicateLibraryImport:      True #Default severity Warning
    DuplicatePackageUsage:       True #Default severity Warning
    DeprecatedPackages:          True #Default severity Warning
    ImplicitLibraries:           True #Default severity Warning
    DisconnectedPorts:           True #Default severity Critical
    IntNoRange:                  True #Default severity Warning
```

The most interesting part is that it is possible to configure which kind of diagnostics should be displayed, and their severity levels.

Schottkyc137:
This relates to #282 and is based upon the branch from #282

Closes #162 

While not all features requested in the Issue are implemented here, the main one is. I prefer to have separate issues that are better contained, so I urge anyone interested to open new issues

1) Assigns the severity level to the Error Code
2) Allows to define the severity level using the configuration:

Todos:
- [x] Automatically re-evaluate the diagnostics if the configuration has changed.

example `vhdl_ls.toml`:
```toml
[libraries]
lib.files = [ "..."]

[lint]
unused = 'error'
misplaced_attribute_spec = false # Disable the misplaced_attribute_spec diagnostic
```


Schottkyc137:
@kraigher (and anyone who is interested): Would you prefer the `error_codes` dictionary to be default configuration as is shown in the example above or in a separate `[config]` dictionary together with future configurations?

kraigher:
I think you need to consider who would disable errors and why. 

Errors that come from the VHDL standard should almost never be reasonable to disable or downgrade to warning. Most VHDL-LS diagnostics  errors are or this type.

The use cases I see are:

1. relaxed rules such as use of shared variables. 

2. to disable false positive errors in VHDL_LS. But for this you want to disable it next to the offending line and not a global configuration file. Ideally users would make bug reports. A design goal of VHDL-LS was to never produce false positives but be silent when it cannot be sure.

3. To configure optional lints, things like warning for variable not used etc. This would also likely be locally per line and not globally.

4. If there are more "lints" added like the VHDL Style Guide where there are rules not required by the standard but rather by team convention. These would make sense globally configure. This could be warning for latches etc.

Since the most useful usecases for configuration of diagnostics do not involve language errors I think calling it "error_codes" is not the best name. I would call it "lint" or "diagnostics", this would be more proper name for future lints that are optional and not language errors violating the LRM.

Schottkyc137:
I agree. I have changed the table name to `lint` and I have added a section in the readme to signify that we generally discourage overwriting syntax or analysis errors.

PR #284, opened 2024-04-07
Add support for different VHDL standards

Schottkyc137:
This adds the initial infrastructure to support different VHDL standards.

Currently, the standards `1993`, `2008` and `2019` are implemented.

- [x] Enable specifying the default standard to analyse in the config
  Example:
  ```toml
  standard = "1993"
  [libraries]
  ...
  ```
- [x] Use different keywords and attributes based on that config


Schottkyc137:
Per-file and per-library overwrites are planned too.
I'm not entirely sure how to implement this though as the parser is the same for all files at the moment. A per-file parser seems a little inefficient and cumbersome. Maybe a per-file parser with reference to symbol tables is best here, allowing multiple files to share the same symbol table.

Schottkyc137:
@kraigher there is an `AttributeDesignator::Descending` attribute defined but I couldn't find it in the standard (regardless of the version). Are you sure this exists?

kraigher:
If you cannot find it in the LRM it was probably added by mistake.

Issue #296, opened 2024-04-22
Adding automatting


KarelPeeters:
## Intro

I'm investigating how hard it would be to add automating as a feature, as part of https://github.com/VHDL-LS/rust_hdl/issues/156. This project already has a parser, now we "only" need to add the other direction :slightly_smiling_face:.

Inspiration can be taken from rustfmt, notably [`Design.md`](https://github.com/rust-lang/rustfmt/blob/master/Design.md) and the more up-to-date [`Contributing.md`](https://github.com/rust-lang/rustfmt/blob/master/Contributing.md).

## Current state

I noticed the file [`display.rs`](https://github.com/VHDL-LS/rust_hdl/blob/master/vhdl_lang/src/ast/display.rs) has implementations for some, but not all, AST nodes. As far as I can tell statements and anything that would require them (up to and including `AnyDesignUnit::Secondary`) does not yet have a `Display` implementation.
Is the intention that this gets extended to cover the entire AST?

Some limitations of the current design:
* Indentation is hard-coded per AST node type, which does not extend to arbitrarily nested expressions.
* Parenthesis insertion is decided based on precedence levels, potentially overriding user intent.
* Comments are dropped, these need to be preserved.
* There is no line wrapping, and it's not obvious how to add this.

## Future

Some decisions need to be made:
* Should the formatter be whitespace only or more all-encompassing (including line formatting and blank lines) like rustfmt?
   * I think white-space only is much easier to implement, since we don't need to make any complicated line-breaking decisions.
* How much information do we want to add to the AST to enable formatting? For example, comments can be handled in two ways: either add them to the AST and then emit them again, or keep them in a separate data structure and re-insert them during stringifying. The same is true for line breaks.

I'll do some more research and maybe start on a prototype, I just wanted to start the discussion as early as possible.

kraigher:
First of all the Display trait is a dead end for formatting. To do proper formatting you need to have a context such as the current indentation level and probably also settings. 

> How much information do we want to add to the AST to enable formatting? For example, comments can be handled in two ways: either add them to the AST and then emit them again, or keep them in a separate data structure and re-insert them during stringifying. The same is true for line breaks.

The current architecture has been to not include comments in the AST but rather have each AST be aware of the span of Tokens it consists of. The comments can be extracted from the token span and be kept in this way. Each Token is associated with either leading or trailing comments. The comments are stored in the tokens and are not tokens themselves.

[...]

Issue #158, opened 2023-02-21
[Bug] Format of `textDocument/documentSymbol` LSP response doesn't allegedly match the LSP specification


pidgeon777:
[Neovim](https://neovim.io/) is a text editor tool which makes use of the Language Server Protocol (LSP) to provide code completion, diagnostics, and other features. The LSP specification defines a response format for the `textDocument/documentSymbol` request, and the one generated by `rust_hdl` seems not be be 100% compliant to the specification.

For example, let consider the following code:

```vhdl
library ieee;
use ieee.std_logic_1164.all;

entity ENTITY_TOP is
  generic (
    GEN : integer := 0
  );
  port (
    INP : in std_logic
  );
end entity;

architecture arch of ENTITY_TOP is
  signal sig : std_logic := '0';

  component ENTITY_1
    generic (
      GEN : integer := 0
    );
    port (
      INP : in std_logic
    );
  end component;

  component ENTITY_2
    generic (
      GEN : integer := 0
    );
    port (
      INP : in std_logic
    );
  end component;

begin

  ENTITY_1_1 : entity work.ENTITY_1(arch1)
  generic map(
    GEN => GEN
  )
  port map(
    INP => INP
  );

  ENTITY_1_2 : entity work.ENTITY_1(arch2)
  generic map(
    GEN => GEN
  )
  port map(
    INP => INP
  );

  ENTITY_2_1 : entity work.ENTITY_2(arch1)
  generic map(
    GEN => GEN
  )
  port map(
    INP => INP
  );

  ENTITY_2_2 : entity work.ENTITY_2(arch2)
  generic map(
    GEN => GEN
  )
  port map(
    INP => INP
  );

  PROC_p: process(INP)
  -----------------------------
  variable var_v : integer := 0;
  -----------------------------
  begin
      -----------------------------
      var_v := 0;
      -----------------------------
      if (INP = '1') then
        sig <= '1';
      else
        sig <= '0';
      end if;
      -----------------------------
  end process;
  -----------------------------

  process
  -----------------------------
  variable var_v : integer := 0;
  -----------------------------
  begin
      -----------------------------
      var_v := 0;
      -----------------------------
      if (INP = '1') then
        sig <= '1';
      else
        sig <= '0';
      end if;
      -----------------------------
  end process;
  -----------------------------
end architecture;
```

In my test, I compared the `textDocument/documentSymbol` response of the following LSP servers:

- `rust_hdl`
[...]

PR #301, opened 2024-05-04
Replace `WithPos<T>` with `WithTokenSpan<T>`

Schottkyc137:
This is the first step towards #296 and helps towards #158 

Previously, the struct `WithPos` encompassed a generic item and the position in the source code where that item belongs to.
With this PR, the implementation has been changed to facilitate the `WithTokenSpan` struct, replacing the source code information with the tokens that make up that item.
Using a `TokenAccess` context object, one can then retrieve information about the source-code location, in addition to other information like comments arched to that token.

The only thing missing from this PR is a needed rework of the testing framework. With the token information now being attached to lots of elements of the AST, a lot of tests no longer work properly.


pidgeon777:
This LSP could be used as a good reference, for scope implementations and many other features:

https://github.com/vhdl-linter/vhdl-linter

@Schottkyc137 did you already know it?

Schottkyc137:
> This LSP could be used as a good reference, for scope implementations and many other features:
> 
> https://github.com/vhdl-linter/vhdl-linter
> 
> @Schottkyc137 did you already know it?

I have seen that you mentioned it a couple of times. Not sure if they do anything special, but from my perspective it's just a 'correct' implementation of the LSP, attaching declaration position and source position to every element. Is there more to it?

Issue #225, opened 2023-11-18
Missing syntactical elements


Schottkyc137:
The following is a list of elements that are currently not syntactically analyzed. Mostly this is because these elements are rarely used. However, for full compliance with the standard, these should be analyzed eventually.
Some of these can also be a starting point for new developers as these features (e.g., groups) are usually small and self-contained.
Below are also minimal reproducible examples that do not parse as of today.

This is also a good place to request prioritization on a feature if this feature is needed in a codebase.
- [ ] Package declaration inside a declarative part
  ```vhdl
  entity ent is
  end entity;

  architecture arch of ent is
    package my_pkg is
        -- ...
     end package;
  begin
  end arch;
  ```
- [ ] Package body inside a declarative part
  ```vhdl
  entity ent is
  end entity;

  architecture arch of ent is
    package my_pkg is
        -- ...
     end my_pkg;
    package body my_pkg is
        -- ...
     end package body;
  begin
  end arch;
  ```
- [ ] Disconnect Specification
  ```vhdl
  disconnect sig after 100 ns;
  ```
- [ ] Missing keywords and context where they are used
  - [ ] Sequence
  - [ ] Property 
- [ ] Group declaration
  ```vhdl
  group my_group : template_name ( A, B ) ;
  ```
- [ ] Group template declaration
  ```vhdl
  group template_name is ( sequence, function ) ;
  ```
- [ ] Concurrent select guarded and force assignment
   ```vhdl
   with a select b <= force c when d, e when f;
   with a select b <= guarded c when d, e when f;
   ```
- [ ] Question mark after select
   ```vhdl
   with a select? b <= c when d, e when f;
   ```
- [ ] Release assignment
   ```vhdl
   a <= release out;
   ```

Chris44442:
Hi, I haven't worked with the code base before. I have tried fixing the first of your issues [here](https://github.com/Chris44442/rust_hdl/tree/add_declarative_package_declaration)

I didn't understand what to do in some parts, e.g. in the analyze_declaration function or the HasEntityId impl. I have marked it with TODO. Either way your example code parses now and the package is registered as a LRM 4.7 Package declaration, instead of a LRM 4.9 Package instatiation declaration.

Should I make a PR?

Schottkyc137:
> Hi, I haven't worked with the code base before. I have tried fixing the first of your issues [here](https://github.com/Chris44442/rust_hdl/tree/add_declarative_package_declaration)
> 
[...]

PR #307, opened 2024-05-30
Add package declaration in declarative part

Chris44442:
This is addressing the first two tasks of #225

### DIFF
--- a/vhdl_lang/src/analysis/declarative.rs
+++ b/vhdl_lang/src/analysis/declarative.rs
@@ -614,11 +614,11 @@ impl<'a, 't> AnalyzeContext<'a, 't> {
                     scope.add(ent, diagnostics);
                 }
             }
-            Declaration::PackageDeclaration(..) => {
-                // TODO
+            Declaration::PackageDeclaration(ref mut unit) => {
+                self.analyze_package(unit, diagnostics)?;
             }
-            Declaration::PackageBody(..) => {
-                // TODO
+            Declaration::PackageBody(ref mut unit) => {
+                self.analyze_package_body(unit, diagnostics)?;
             }
             Declaration::Configuration(..) => {}
             Declaration::View(view) => {
--- a/vhdl_lang/src/analysis/declarative.rs
+++ b/vhdl_lang/src/analysis/declarative.rs
@@ -42,6 +42,7 @@ impl Declaration {
                     | Use(_)
                     | Package(_)
                     | PackageDeclaration(_)
+                    | PackageBody(_)
                     | Configuration(_)
                     | View(_)
             ),
@@ -63,6 +64,7 @@ impl Declaration {
                     | Use(_)
                     | Package(_)
                     | PackageDeclaration(_)
+                    | PackageBody(_)
                     | View(_)
             ),
             // LRM: package_body_declarative_item
@@ -89,6 +91,7 @@ impl Declaration {
                     | Use(_)
                     | Package(_)
                     | PackageDeclaration(_)
+                    | PackageBody(_)
             ),
             // LRM: package_declarative_item
             AnyEntKind::Design(Design::Package(..)) => matches!(
@@ -104,6 +107,7 @@ impl Declaration {
                     | Use(_)
                     | Package(_)
                     | PackageDeclaration(_)
+                    | PackageBody(_)
                     | View(_)
             ),
             _ => {
@@ -611,6 +615,9 @@ impl<'a, 't> AnalyzeContext<'a, 't> {
                 }
             }
             Declaration::PackageDeclaration(..) => {
+                // TODO
+            }
+            Declaration::PackageBody(..) => {
                 // TODO
             }
             Declaration::Configuration(..) => {}
--- a/vhdl_lang/src/analysis/declarative.rs
+++ b/vhdl_lang/src/analysis/declarative.rs
@@ -626,12 +626,33 @@ impl<'a, 't> AnalyzeContext<'a, 't> {
             Declaration::Configuration(..) => {}
             Declaration::Disconnection(ref mut disc) => {
                 let DisconnectionSpecification {
-                    ident: _,
+                    ident,
                     subtype_indication,
                     expression,
                 } = disc;
                 self.expr_with_ttyp(scope, self.time(), expression, diagnostics)?;
                 self.analyze_subtype_indication(scope, subtype_indication, diagnostics)?;
+
+                let subtype = as_fatal(self.resolve_subtype_indication(
+                    scope,
+                    subtype_indication,
+                    diagnostics,
+                ));
+                if let Ok(Some(subtype)) = subtype {
+                    if let GuardedSignalList::Ident(ref mut ident) = ident {
+                        scope.add(
+                            self.arena.define(
+                                self.ctx,
+                                ident,
+                                parent,
+                                AnyEntKind::Disconnection(subtype),
+                                src_span,
+                                Some(self.source()),
+                            ),
+                            diagnostics,
+                        );
+                    }
+                }
             }
             Declaration::View(view) => {
                 if let Some(view) = as_fatal(self.analyze_view_declaration(
@@ -1210,6 +1231,7 @@ fn get_entity_class(ent: EntRef) -> Option<EntityClass> {
             Design::Context(_) => None,
         },
         AnyEntKind::View(_) => None,
+        AnyEntKind::Disconnection(_) => None,
     }
 }
 
--- a/vhdl_lang/src/analysis/declarative.rs
+++ b/vhdl_lang/src/analysis/declarative.rs
@@ -44,6 +44,7 @@ impl Declaration {
                     | PackageDeclaration(_)
                     | PackageBody(_)
                     | Configuration(_)
+                    | Disconnection(_)
                     | View(_)
             ),
             // LRM: configuration_declarative_item
@@ -65,6 +66,7 @@ impl Declaration {
                     | Package(_)
                     | PackageDeclaration(_)
                     | PackageBody(_)
+                    | Disconnection(_)
                     | View(_)
             ),
             // LRM: package_body_declarative_item
@@ -76,6 +78,7 @@ impl Declaration {
                 | Overloaded::UninstSubprogram(..),
             )
             | AnyEntKind::Concurrent(Some(Concurrent::Process))
+            // LRM protected type body declarative item
             | AnyEntKind::Type(named_entity::Type::Protected(..)) => matches!(
                 self,
                 Object(ObjectDeclaration {
@@ -107,7 +110,7 @@ impl Declaration {
                     | Use(_)
                     | Package(_)
                     | PackageDeclaration(_)
-                    | PackageBody(_)
+                    | Disconnection(_)
                     | View(_)
             ),
             _ => {
@@ -621,6 +624,7 @@ impl<'a, 't> AnalyzeContext<'a, 't> {
                 self.analyze_package_body(unit, diagnostics)?;
             }
             Declaration::Configuration(..) => {}
+            Declaration::Disconnection(..) => {} // @TODO
             Declaration::View(view) => {
                 if let Some(view) = as_fatal(self.analyze_view_declaration(
                     scope,
--- a/vhdl_lang/src/analysis/declarative.rs
+++ b/vhdl_lang/src/analysis/declarative.rs
@@ -624,7 +624,15 @@ impl<'a, 't> AnalyzeContext<'a, 't> {
                 self.analyze_package_body(unit, diagnostics)?;
             }
             Declaration::Configuration(..) => {}
-            Declaration::Disconnection(..) => {} // @TODO
+            Declaration::Disconnection(ref mut disc) => {
+                let DisconnectionSpecification {
+                    ident: _,
+                    subtype_indication,
+                    expression,
+                } = disc;
+                self.expr_with_ttyp(scope, self.time(), expression, diagnostics)?;
+                self.analyze_subtype_indication(scope, subtype_indication, diagnostics)?;
+            }
             Declaration::View(view) => {
                 if let Some(view) = as_fatal(self.analyze_view_declaration(
                     scope,
--- a/vhdl_lang/src/analysis/declarative.rs
+++ b/vhdl_lang/src/analysis/declarative.rs
@@ -41,6 +41,7 @@ impl Declaration {
                     | SubprogramBody(_)
                     | Use(_)
                     | Package(_)
+                    | PackageDeclaration(_)
                     | Configuration(_)
                     | View(_)
             ),
@@ -61,6 +62,7 @@ impl Declaration {
                     | SubprogramBody(_)
                     | Use(_)
                     | Package(_)
+                    | PackageDeclaration(_)
                     | View(_)
             ),
             // LRM: package_body_declarative_item
@@ -86,6 +88,7 @@ impl Declaration {
                     | SubprogramBody(_)
                     | Use(_)
                     | Package(_)
+                    | PackageDeclaration(_)
             ),
             // LRM: package_declarative_item
             AnyEntKind::Design(Design::Package(..)) => matches!(
@@ -100,6 +103,7 @@ impl Declaration {
                     | SubprogramInstantiation(_)
                     | Use(_)
                     | Package(_)
+                    | PackageDeclaration(_)
                     | View(_)
             ),
             _ => {
@@ -605,6 +609,9 @@ impl<'a, 't> AnalyzeContext<'a, 't> {
                     }
                     scope.add(ent, diagnostics);
                 }
+            }
+            Declaration::PackageDeclaration(..) => {
+                // TODO
             }
             Declaration::Configuration(..) => {}
             Declaration::View(view) => {
--- a/vhdl_lang/src/analysis/names.rs
+++ b/vhdl_lang/src/analysis/names.rs
@@ -188,6 +188,7 @@ impl<'a> ResolvedName<'a> {
             }),
             AnyEntKind::Type(_) => ResolvedName::Type(TypeEnt::from_any(ent).unwrap()),
             AnyEntKind::View(_) => ResolvedName::Final(ent),
+            AnyEntKind::Disconnection(_) => ResolvedName::Final(ent),
             AnyEntKind::Overloaded(_) => {
                 return Err((
                     "Internal error. Unreachable as overloaded is handled outside".to_owned(),
@@ -258,6 +259,7 @@ impl<'a> ResolvedName<'a> {
             }
             AnyEntKind::File(_)
             | AnyEntKind::View(_)
+            | AnyEntKind::Disconnection(_)
             | AnyEntKind::InterfaceFile(_)
             | AnyEntKind::Component(_)
             | AnyEntKind::Concurrent(_)
--- a/vhdl_lang/src/analysis/names.rs
+++ b/vhdl_lang/src/analysis/names.rs
@@ -1848,6 +1848,7 @@ impl Declaration {
             Declaration::PackageDeclaration(_) => "package",
             Declaration::PackageBody(_) => "package body",
             Declaration::Configuration(_) => "configuration",
+            Declaration::Disconnection(_) => "disconnection",
             Declaration::View(_) => "view",
         }
     }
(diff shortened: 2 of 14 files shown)


Chris44442:
@Schottkyc137 

This is what I came up with. I am not sure if the changes in analysis/design_unit.rs  are sensible. I am still not quite 100% what is going on in that particular code region tbh lol. Oh and I only added one test. Not sure if we need more.

Schottkyc137:
So far things look good to me. One thing to be wary of is that the package declaration might not be visible in the global scope. In other words,
```
package foo is
begin
  -- ...
end package foo;
```

Would mean that `foo` can be imported globally (I.e. `use foo.bar` would make `bar` visible) whereas
```
entity baz

package foo is
begin
  -- ...
end package foo;

end baz;
```

would probably mean that `foo` can only be accessed from within `baz`. I'm unsure what the LRM says in this regard though.

Also, I think that we should have more tests. The main purpose of the thorough testing is for regressions, I.e. checking that things also work after auxiliary changes.

I'm somewhat busy at the moment, so I can only give any further, code specific review later; latest however Sunday.

That being said thanks again for taking up this issue!

Chris44442:
Alright thanks, I will have a look at your notes later this day.

I have also pushed the Disconnection Specification to my fork. It now parses, but analysing Disconnect still does nothing for now. I will try to figure out how the analyse stuff works next, so I can try to implement it. The thing with Disconnect though is that it also depends on Guarded, which only seems partially implemented at the moment.

vhdl_lang/src/analysis/declarative.rs
```diff
@@ -606,7 +617,14 @@ impl<'a, 't> AnalyzeContext<'a, 't> {
...
                 }
             }
+            Declaration::PackageDeclaration(ref mut unit) => {
+                self.analyze_package(unit, diagnostics)?;
+            }
+            Declaration::PackageBody(ref mut unit) => {
+                self.analyze_package_body(unit, diagnostics)?;
+            }
             Declaration::Configuration(..) => {}
+            Declaration::Disconnection(..) => {} // @TODO
```
Schottkyc137, line 627: Probably it's out of scope for this PR to start analysing Disconnect and guarded statements thoroughly. But some limited analysis could still be done. For example, for the statement
```
disconnect foo, bar : resolved_bit after 5 ns ;
                    ^              ^                             ^
                    |                |                               Check that this is an expression of type time
                    |                Check that this is a valid type that matches the type declared by the single names
                   Check that this is a valid signal name
```

For all of this, examples should exist somewhere in the code on how to analyze this.
- For the 'expect that something is an expression of type time' part, you can have a look at [how `wait` statements are analysed](https://github.com/VHDL-LS/rust_hdl/blob/805ab71a5ce8ce7af30557dabc20728b0205c68a/vhdl_lang/src/analysis/sequential.rs#L159)
- To check the type and the signal names, you could look at [how `alias` declarations are analysed](https://github.com/VHDL-LS/rust_hdl/blob/805ab71a5ce8ce7af30557dabc20728b0205c68a/vhdl_lang/src/analysis/declarative.rs#L243), though this is not as straightforward as the time expression.

vhdl_lang/src/analysis/declarative.rs
```diff
@@ -606,7 +617,14 @@ impl<'a, 't> AnalyzeContext<'a, 't> {
                     scope.add(ent, diagnostics);
                 }
             }
+            Declaration::PackageDeclaration(ref mut unit) => {
+                self.analyze_package(unit, diagnostics)?;
+            }
+            Declaration::PackageBody(ref mut unit) => {
+                self.analyze_package_body(unit, diagnostics)?;
```
Schottkyc137, line 624: This (and the package declaration analysis) is fine, but I think that this declares them in the global namespace. As I have written in the second comment, check that this is correct.

vhdl_lang/src/analysis/names.rs
```diff
@@ -1845,7 +1845,10 @@ impl Declaration {
             Declaration::SubprogramBody(_) => "subprogram body",
             Declaration::Use(_) => "use",
             Declaration::Package(_) => "package instantiation",
+            Declaration::PackageDeclaration(_) => "package",
+            Declaration::PackageBody(_) => "package body",
             Declaration::Configuration(_) => "configuration",
+            Declaration::Disconnection(_) => "disconnection",
```
Schottkyc137, line 1851: I think calling this 'disconnection' as opposed to 'disconnect' could cause confusion. The rest of the names follow the names also given in the LRM, I think we should therefore call this 'disconnect' as well.

vhdl_lang/src/ast.rs
```diff
@@ -1375,6 +1378,10 @@ pub struct ConfigurationSpecification {
     pub vunit_bind_inds: Vec<VUnitBindingIndication>,
 }
 
+/// LRM 7.4 Disconnection specification
+#[derive(PartialEq, Debug, Clone)]
+pub struct DisconnectionSpecification {}
```
Schottkyc137, line 1383: Here, we should put all of the AST elements. You are currently parsing this very nicely, but then drop all of the information. This should include all relevant AST elements, such as 
- The ident list
- The type
- The wait expression

vhdl_lang/src/ast/search.rs
```diff
@@ -1075,14 +1075,21 @@ impl Search for Declaration {
...
+                return_if_found!(decl.search(ctx, searcher));
+            }
+            Declaration::PackageBody(ref body) => {
+                return_if_found!(body.search(ctx, searcher));
+            }
             Declaration::Configuration(_) => {
                 // @TODO
             }
+            Declaration::Disconnection(_) => {
+                // @TODO
```
Schottkyc137, line 1091: This is used, for example, by the `find all references` functionality of the LSP. Here, you can just delegate the search method to each children of the `Disconnect` AST element. The special search methods in `ident`s or `expression`s take care of the rest.

vhdl_lang/src/named_entity.rs
```diff
@@ -631,8 +632,11 @@ impl HasEntityId for Declaration {
             Declaration::SubprogramBody(body) => body.ent_id(),
             Declaration::SubprogramInstantiation(decl) => decl.ent_id(),
             Declaration::Package(pkg) => pkg.ent_id(),
+            Declaration::PackageDeclaration(pkg) => pkg.ent_id(),
+            Declaration::PackageBody(pkg) => pkg.ent_id(),
             Declaration::Use(_) => None,
             Declaration::Configuration(_) => None,
+            Declaration::Disconnection(_) => None, // TODO
```
Schottkyc137, line 639: This would just return the `EntityId` of the disconnect specification. For simplicity, you can just implement it in a way that for an identifier list > 1, i.e., `disconnect a, b, c, ...`, a new `Disconnect` struct should be created.

vhdl_lang/src/syntax/declarative_part.rs
```diff
@@ -332,4 +355,16 @@ var not_a_var: broken;
...
+    #[test]
+    fn parse_declarative_part_with_disconnection() {
+        let code = Code::new(
+            "\
+constant foo: real := 5.1;
+disconnect my_signal : integer after 42 ms;
+signal bar: std_logic;
+",
+        );
+        code.with_stream_no_diagnostics(parse_declarative_part);
```
Schottkyc137, line 368: This is a good test, but it doesn't do a lot yet. It would be better to also check the value of each declaration and see that it matches.

vhdl_lang/src/syntax/disconnection.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::ast::*;
+use vhdl_lang::syntax::parser::ParsingContext;
+
+/// LRM 7.4 Disconnection Specification
+pub fn parse_disconnection_specification(
+    ctx: &mut ParsingContext<'_>,
+) -> ParseResult<WithTokenSpan<DisconnectionSpecification>> {
+    let start_token = ctx.stream.expect_kind(Disconnect)?;
+    if ctx.stream.next_kind_is(Identifier) {
+        let _ = parse_ident_list(ctx);
```
Schottkyc137, line 22: As I said in a previous comment, this information should be kept in the `DisconnectSpecification` struct.

I really like the separation of concerns in this file though, i.e., that you created a separate file for the purposes of parsing the disconnect statement.

vhdl_lang/src/syntax/disconnection.rs
```diff
@@ -0,0 +1,95 @@
...
+    let _ = parse_expression(ctx);
+    let end_token = ctx.stream.expect_kind(SemiColon)?;
+    Ok(WithTokenSpan::new(
+        DisconnectionSpecification {},
+        TokenSpan::new(start_token, end_token),
+    ))
+}
+
+#[cfg(test)]
+mod tests {
```
Schottkyc137, line 40: These tests are very good and exactly what we want. They check for correctness and, more importantly, avoid regressions.
In addition to these tests, you could optionally also test cases that are expected to fail, i.e.
```vhdl
disconnect S: T before 42 ms;
```

Generally, I think that this is not super important here but this helps in justifying that we show good diagnostics, and that we don't let code pass that is clearly wrong.

Schottkyc137 (commented):
All of this looks really good so far.
I have made comments where you had TODO marks and where I think some improvements could be made.

vhdl_lang/src/analysis/names.rs
Chris44442: The keyword is "disconnect" but LRM 7.4 is called "Disconnection Specification", similar to e.g. "Configuration Specification". Anyway I don't care which tbh.

Schottkyc137: Ah, I didn't realize this. In this case, I think we should stick to the LRM like you proposed.

vhdl_lang/src/ast.rs
Chris44442: This should be fixed now. The only thing I couldn't do was multiple idents. I left a little TODO for that. I am also not quite sure what ident (if any) to return when it says Others or All. Maybe this needs to be done in the analyzer or so.

vhdl_lang/src/syntax/design_unit.rs
```diff
@@ -933,4 +933,40 @@ end entity y;
...
+        assert_eq!(tokens[9].kind, Identifier);
+        assert_eq!(tokens[10].kind, SemiColon);
+        assert_eq!(tokens[11].kind, Package);
+        assert_eq!(tokens[12].kind, Body);
+        assert_eq!(tokens[13].kind, Identifier);
+        assert_eq!(tokens[14].kind, Is);
+        assert_eq!(tokens[15].kind, End);
+        assert_eq!(tokens[16].kind, Package);
+        assert_eq!(tokens[17].kind, Body);
+    }
```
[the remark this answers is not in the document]
Chris44442: I think I need a little help here. I have tried to do what you said but I find it very hard to fill a default ArchitectureBody with the necessary values. The other tests use this simple_architecture, but I don't think this helps me much. I have put the test I thought of in comments.

vhdl_lang/src/syntax/declarative_part.rs
Chris44442: I have moved this over into Disconnetion.rs, where I tried to do a lot more in depths tests.

vhdl_lang/src/syntax/disconnection.rs
Chris44442: Done, see comment above.

vhdl_lang/src/syntax/disconnection.rs
Chris44442: Done. I have added a few more test cases that fail. I had a little problem, because some of these error-test-cases have no diagnostics. In these cases I simply asserted if the declarations had any errors.

vhdl_lang/src/ast/search.rs
Chris44442: I have done it. But it doesn't work for some reason.

When I test it in:

```vhdl
library ieee;
use ieee.std_logic_1164.all;

entity ent is
end entity;

architecture arch of ent is
  signal foo,bar : std_logic := '0';
  constant abc : time := 10 ms;
  disconnect all : std_logic after abc;

begin

  foo <= bar after abc;

end arch;
```

... the "abc" is not listed when I use LSP goto references. Not quite sure why. Maybe because it must be a time_expression, not just a generic expression, mh.

vhdl_lang/src/analysis/declarative.rs
Chris44442: I did the type checking and also the expression checking and also added some tests for that. Pretty cool that it works just like that. Anyway I have no clue how to check for the names of the identifier tbh.

vhdl_lang/src/named_entity.rs
Chris44442: Done. But I still haven't figured out how to parse multiple idents, so there will be more to do here once I have.

vhdl_lang/src/analysis/tests/declarations.rs
```diff
@@ -100,3 +100,96 @@ end entity test;
...
+entity ent is
+end entity;
+architecture arch of ent is
+  constant foo : time := 10 ns;
+  disconnect all : integer after foo;
+begin
+end arch;
+    ",
+    );
+    check_diagnostics(builder.analyze(), vec![])
```
Schottkyc137, line 119: This could be simplified by using `check_no_diagnostics(builder.analyze())`

vhdl_lang/src/syntax/disconnection.rs
Schottkyc137: Nice!

vhdl_lang/src/ast.rs
```diff
@@ -1375,6 +1378,14 @@ pub struct ConfigurationSpecification {
     pub vunit_bind_inds: Vec<VUnitBindingIndication>,
 }
 
+/// LRM 7.4 Disconnection specification
+#[derive(PartialEq, Debug, Clone)]
+pub struct DisconnectionSpecification {
+    pub ident: Option<WithDecl<Ident>>,
```
Schottkyc137, line 1384: The way I would solve the ident problem is twofold.
First, declare an enum
```rust
pub enum DisconnectSignalSpecification {
    All,
    Others,
    Ident(WithDecl<Ident>)
}
```

And then change the disconnect specification ident to be an instance of that enum, i.e.,
```rust
pub struct DisconnectionSpecification {
    pub ident : DisconnectSignalSpecification,
    // ...
}
```

Then, to allow for multiple idents, just emit multiple specification, i.e.,
```vhdl
disconnect a, b, c : some_type after 10 ns;
```
would declare three `DisconnectSpecification`s, each with a different ident but with the same auxiliary information (i.e., the subtype and the time expression stays the same). This is also the way it's done for, for example, signal declarations.

vhdl_lang/src/syntax/design_unit.rs
```diff
@@ -933,4 +933,34 @@ end entity y;
...
+    //",
+    //        );
+    //
+    //assert_eq!(
+    //    code.with_stream_no_diagnostics(parse_architecture_body()),
+    //        ArchitectureBody {
+    //            span: code.token_span(),
+    //            context_clause: ContextClause::default(),
+    //            ident: code.s1("arch").decl_ident(),
+    //            entity_name: None, // TODO
```
Schottkyc137, line 959: This is just the name of the entity, i.e. `ent` in this example. You should simply be able to get a reference to that using `WithRef::new(code.s1("ent").ident())`

vhdl_lang/src/syntax/disconnection.rs
```diff
@@ -0,0 +1,231 @@
...
+pub fn parse_disconnection_specification(
+    ctx: &mut ParsingContext<'_>,
+) -> ParseResult<WithTokenSpan<DisconnectionSpecification>> {
+    let start_token = ctx.stream.expect_kind(Disconnect)?;
+    let mut ident = None;
+
+    if let Some(token) = ctx.stream.peek() {
+        match token.kind {
+            Identifier => {
+                // @TODO allow more than 1 ident
```
Schottkyc137, line 28: To allow for more than one ident, you could simply use a loop, i.e.,
```rust
idents = Vec::new();
loop {
    idents.push(WithDecl::new(ctx.stream.expect_ident()?));
    if (!ctx.stream.next_kind_is(Comma)) {
        break;
    }
}
```

Keeping track of the choice (i.e., whether it's an identifier, `all` or `others` is going to be a bit annoying, but I'm confident you'll find a solution :)

vhdl_lang/src/ast/search.rs
Schottkyc137: I can try to look at your branch later to understand what the problem could be. But there is nothing immediate where I see any issues.

Schottkyc137: Ah, I think this will only work if you analyze the ident

vhdl_lang/src/analysis/tests/declarations.rs
Chris44442: Done

vhdl_lang/src/ast/search.rs
Chris44442: ok, I will try

vhdl_lang/src/syntax/disconnection.rs
Chris44442: After a long back and forth I think I finally got it.

vhdl_lang/src/named_entity.rs
Chris44442: I did the multiple idents, so I think this is should be done now too, right?

vhdl_lang/src/ast.rs
Chris44442: Done. It seems to be working.

vhdl_lang/src/ast/search.rs
Chris44442: I don't know how but it seems I've got it. Although I must admit I tried to mimic what the other items did without fully understanding every step on the way. Better check if I didn't make an obvious mistake. Either way, my example above now works, very cool.

vhdl_lang/src/ast.rs
Chris44442: Done, see https://github.com/VHDL-LS/rust_hdl/pull/307#discussion_r1625150504

vhdl_lang/src/analysis/declarative.rs
Chris44442: I will try to resolve this next tomorrow or so. It is not exactly easy to interpret what the LRM really means, I have to say lol.

vhdl_lang/src/analysis/declarative.rs
Chris44442: Done, see https://github.com/VHDL-LS/rust_hdl/pull/307#discussion_r1625150504

vhdl_lang/src/analysis/declarative.rs
Schottkyc137: Yea, one of the hardest parts of developing a VHDL language server is interpreting the LRM :P

vhdl_lang/src/ast/search.rs
```diff
@@ -1075,14 +1075,27 @@ impl Search for Declaration {
...
+            }
+            Declaration::PackageBody(ref body) => {
+                return_if_found!(body.search(ctx, searcher));
+            }
             Declaration::Configuration(_) => {
                 // @TODO
             }
+            Declaration::Disconnection(disconnect) => {
+                let DisconnectionSpecification {
+                    ident: _,
```
Schottkyc137, line 1092: You could now also search the `ident` here

vhdl_lang/src/named_entity.rs
Schottkyc137: Yep, this looks nice!

vhdl_lang/src/syntax/disconnection.rs
Schottkyc137: This looks really clean now. Good job!

vhdl_lang/src/ast/search.rs
Schottkyc137: All of what you have done looks really good to me. No obvious issues found.

vhdl_lang/src/ast/search.rs
Chris44442: Done. Searching the references of the ident(s) is seems to be working now.

PR #308, opened 2024-06-03
Add support for textDocument/documentHighlight request

rapgenic:
Hi, thank you all for the great project, which has helped me a lot while writing VHDL code!

In most of the other language servers, I use a lot the `documentHighlight` feature, which highlights all the occurrences of a symbol in the current file. Noticing that it was missing from this LS I attempted my own implementation

See the result here:

https://github.com/VHDL-LS/rust_hdl/assets/10487071/ebb064ce-89f1-459c-a37b-edf11d7af977

## Edit

It seems that VScode has some kind of fallback that performs the same result even if the server does not have this capability! (I am actually using another editor - Zed - that would not do the same)

See the before and after in Zed:

### Before

https://github.com/VHDL-LS/rust_hdl/assets/10487071/6162dcfc-965a-4857-86e5-8dc6a060a4f0 

### After

https://github.com/VHDL-LS/rust_hdl/assets/10487071/382ad155-a142-44f8-b731-6372cbd82c85

### DIFF
--- a/vhdl_ls/src/vhdl_server.rs
+++ b/vhdl_ls/src/vhdl_server.rs
@@ -696,17 +696,11 @@ impl VHDLServer {
 
         Some(
             self.project
-                .find_all_references(ent)
+                .find_all_references_in_source(&source, ent)
                 .iter()
-                .filter_map(|pos| {
-                    if pos.source == source {
-                        Some(DocumentHighlight {
-                            range: to_lsp_range(pos.range()),
-                            kind: Some(DocumentHighlightKind::TEXT),
-                        })
-                    } else {
-                        None
-                    }
+                .map(|pos| DocumentHighlight {
+                    range: to_lsp_range(pos.range()),
+                    kind: Some(DocumentHighlightKind::TEXT),
                 })
                 .collect(),
         )
--- a/vhdl_ls/src/vhdl_server.rs
+++ b/vhdl_ls/src/vhdl_server.rs
@@ -182,6 +182,7 @@ impl VHDLServer {
             })),
             workspace_symbol_provider: Some(OneOf::Left(true)),
             document_symbol_provider: Some(OneOf::Left(true)),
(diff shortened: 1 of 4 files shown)


vhdl_ls/src/vhdl_server.rs
```diff
@@ -681,6 +682,36 @@ impl VHDLServer {
...
+        let source = self
+            .project
+            .get_source(&uri_to_file_name(&params.text_document.uri))?;
+
+        let ent = self
+            .project
+            .find_declaration(&source, from_lsp_pos(params.position))?;
+
+        Some(
+            self.project
```
Schottkyc137, line 698: One small optional improvement here:
In your implementation, all references in all files are being searched. After that, only references concerning the current file are kept. This could be improved by just searching in one source file altogether.
For this purpose, the `DesignRoot` has a function `search_source`.
My suggested change is creating a new function (something like `find_all_references_in_source(...)` that only searches in the given source, so the last `filter_map` step could be replaced with a single `map` step.

Schottkyc137 (changes requested):
Thanks a lot for contributing!
This looks good to me so far. I have only added one comment which you can choose to ignore or implement; the choice is up to you.
If you want to ignore it, just mark the conversation as resolved and I will merge your changes.

vhdl_ls/src/vhdl_server.rs
rapgenic: I thought something like that should have existed, but I did not find it, so I went for the simpler but less efficient method. I will have a look at your suggestion and try to implement it.

rapgenic:
What about something like this? I have tried it and it seems to perform actually faster!

Schottkyc137 (approved):

Schottkyc137 (approved):

Schottkyc137:
> What about something like this? I have tried it and it seems to perform actually faster!

Very nice, that's exactly what I had in mind.

PR #320, opened 2024-07-16
AST normalization: store multiple identifiers per declaration instead of multiplying the declaration

Schottkyc137:
For example, the declaration
```vhdl
signal a, b : std_logic;
```
would previously be stores as two separate declarations as if they had been declared like this:
```vhdl
signal a: std_logic;
signal b: std_logic;
```

Now, the AST element contains both identifiers.
This change is mostly for internal purposes. The only noticeable change is that hovers will now produce the former version, instead of the latter where applicable.

### DIFF
--- a/vhdl_lang/src/analysis/analyze.rs
+++ b/vhdl_lang/src/analysis/analyze.rs
@@ -11,6 +11,7 @@ use crate::data::error_codes::ErrorCode;
 use crate::data::*;
 use crate::named_entity::*;
 use crate::syntax::TokenAccess;
+use crate::TokenSpan;
 use fnv::FnvHashSet;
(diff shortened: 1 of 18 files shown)


Schottkyc137:
@kraigher for this change I had to change the implementation of `FoundDeclaration` a bit. Effectively, instead of a single enum, this is now a struct defined as follows:
```rust
pub struct FoundDeclaration<'a> {
    pub reference: &'a Reference,
    pub ast: DeclarationItem<'a>,
}
```
and a declaration with multiple identifiers will emit multiple `FoundDeclaration`s with different idents each, but the same AST element. Do you think, this change is ok? This is the best solutions I came up with for storing the AST as it occurs in the source file.

Issue #40, opened 2019-11-17
Add support for directories or file patterns in vhdl_ls.toml


Bochlin:
Being able to configure the language server to recursively include a directory would simplify project setup. A common practice is to have all RTL sources in a single directory and it should be possible to point to this directory in `vhdl_ls.toml`. Implementing support for directory references would remove the need to update `vhdl_ls.toml` when new files are added to a project.

To account for some variability, it could also be possible to configure include and exclude patterns for the files within the directory.

Suggestion to expand `vhdl_ls.toml` to:

1. Support glob-style patterns in `library.files` (`glob::glob`?)
2. Allow entry in files array to be a directory (checked during config parse) in which case the directory is searched recursively for .vhd files (`**/*.vhd`)
3. Add `excludes` field to library to filter out unwanted files. Supports the same pattern styles as `library.files`. 

Example:
```
[libraries]
lib1.files = [
    'src',
    'tests/*/tb_*.vhd',
]
lib1.exclude = [
    'test/*/*_mock.vhd',
]
```

kraigher:
Some design considerations are:

1
=
Should an error message be given if the pattern does not match any file?
   - If no to the above, what if the pattern is just a plain file name that is missing?


I think an error in case of a missing plain file is definitely required.

2
=
Also regarding the syntax I think it is better with the following:
```toml
[libraries]
lib1.files = [
    'file.vhd',  # Plain file
    'src/*.vhd',  # Pattern without exclude
    {include='tests/*/tb_*.vhd', exclude='*_mock.vhd'}, # Pattern with exclude
]
``` 
The reason being that merged config files can be represented. If two config files both have exclude patterns they could not be merged in your original syntax without the exclude in the second file also affecting the files in the first configuration file. We implicitly merge config files since #39 and I want the merged result to be expressible using the configuration file format. 

Also using this syntax the exclude pattern can be simple because you know it only applies to the include pattern. When using a pattern without exclude it looks the same as with your proposal.

Bochlin:
### 1
If a plain file reference is missing then an error is in place. For an include pattern not matching any files I think that a warning is more than enough and for exclude patterns info may be enough. 
- Plain file missing => Error
- Include pattern without match => Warning
- Exclude pattern without match => Info

 ### 2
I think that both use cases might be valid, however only excluding files from patterns in the "local" configuration is the main use case which makes me lean toward your suggestion.

umarcor:
[...]

PR #370, opened 2025-03-13
Support excluding files in TOML

Xcodo:
Finishes the proposal in #40.

The algorithm is very simple

1. Match all files in the include patterns
2. Match all files in the exclude patterns
3. Remove any files which match an exclude pattern from the include files.

Potentially it would be better to have a layered approach (exact matches first, then wildcard patterns etc.) but I think this is fine for most use cases. It certainly works for the VUnit files.

This is the most rust I've written, so please feel free to alter anything that's not idiomatic.

I've updated the example project TOML, but didn't add wildcards and excludes for PoC as it was going to end up with as many lines as the explicit inclusion!


vhdl_lang/src/config.rs
```diff
@@ -88,7 +90,58 @@ impl LibraryConfig {
             }
         }
         // Remove duplicate file names from the result
-        result.into_iter().unique().collect()
+        result = result.into_iter().unique().collect();
```
Schottkyc137, line 93: I think that, if you are already collecting everything into an iterator, maybe using  a `BTreeSet` or a `HashSet` from the beginning might a better alternative? With the new approach, there is a lot of juggling with vectors and other iterables which adds unwanted complexity.

vhdl_lang/src/config.rs
```diff
@@ -88,7 +90,58 @@ impl LibraryConfig {
...
+                                "Pattern '{stripped_pattern}' did not match any file"
+                            )));
+                        }
+                    }
+                    Err(err) => {
+                        messages.push(Message::error(format!("Invalid pattern '{pattern}' {err}")));
+                    }
+                }
+            }
+        }
```
Schottkyc137, line 139: There is now a lot of code duplication as this code is basically the same as the code that generates the file patterns in the first place. I suggest to factor out the main logic into a separate function  and generate the two file lists from that function.

vhdl_lang/src/config.rs
```diff
@@ -146,6 +199,29 @@ impl Config {
...
+                            .to_str()
+                            .ok_or_else(|| format!("Could not convert {path:?} to string"))?
+                            .to_owned();
+                        exclude_patterns.push(path);
+                    }
+                } else {
+                    return Err(format!("excludes for library {name} is not array"));
+                }
+            }
+
```
Schottkyc137, line 224: I think that, similar to my comment above, there is some code duplication here that could be avoided.

Schottkyc137 (changes requested):
Thanks a lot for contributing this feature! I think the implementation is really solid, but could be improved in one or two points regarding code duplication. Other than that, I really like the simplifications that this change brings to the example project and others like it. It's not often that a Pull request has more deleted lines than lines that were added (which is a good thing!)

Xcodo:
Thanks for the feedback. All your points make sense, I'll try to have a look again in the next few days and simplify it.

vhdl_lang/src/config.rs
Xcodo: I've implemented this with `BTreeSet`, but the test results are no non-deterministic. `config::tests::config_from_str` in particular seems to fail approximately half of the time due to the ordering of the files in the `assert!` on `&[pkg2_path, absolute_vhd]`. The same thing occurs with `HashSet`.

How important is the returned file order for the rest of the VHDL-LS? If it isn't, then I could change the tests to check the contents of the collection without the order, but only if it really doesn't matter.

vhdl_lang/src/config.rs
Xcodo: I've extracted the logic into a function. Would you typically add tests for this new function too, though it's result is tested in the tests for its callers?

vhdl_lang/src/config.rs
Schottkyc137: I don't think the file order plays any role. `vhdl_ls` should is capable of determining the compile order independent on the actual order of files that were passed in, so you can just re-write your test to compare the expected with the actual disregarding the order.

vhdl_lang/src/config.rs
Schottkyc137: I generally wouldn't write a separate test-case for this if it's already tested extensively through other means, so I think this is fine.

vhdl_lang/src/config.rs
Xcodo: Thanks. I've done that and all seems to be good now.

PR #376, opened 2025-04-22
Move information to workspace and specify MSRV

oscargus:
Minimum supported rustc version (MSRV) was not provided. Also, it is more convenient to have shared info in the workspace Cargo.toml.

MSRV is 1.82 for the library (because of `iter::repeat_n`), but the use of brunch 0.9 require 1.85. As this is a test dependency, I decided to downgrade to 0.8 to keep MSRV a bit lower than the latest. Updating to 1.85 is now easier as that will also require updating to edition to 2024, which can be done in a single location.

In general, specifying version with "0" is a source for issues since it is OK to have breaking API changes for minor versions if major version is "0". For major version 1 and higher, there should be no breaking API changes without changing major version. I didn't specify a version for the crates that has "0" though (except for brunch then).  The drawback is that cargo will not automatically update minor versions for major version "0", so this will have to be done manually. Still, it is probably better to update manually rather than breaking it. (Or at least specify that `--locked` should be used when building locally.)

Finally, I also changed the repository to point to the "correct"(?) one (although there is a redirect).

One may consider moving the dependencies to using the workspace as well, but there weren't that many shared dependencies (and since the version specification is pretty wide it may not give that much of a benefit anyway).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,9 +103,9 @@ checksum = "5c8214115b7bf84099f1309324e63141d4c5d7cc26862f97a0a857dbefe165bd"
 
 [[package]]
 name = "brunch"
-version = "0.9.0"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0213369ce5beadc65e9218c969c663c0fd6e730df41d8c02a39ecaf9c5022f70"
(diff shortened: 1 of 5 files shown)


Schottkyc137:
Thanks! Those changes all look good and make sense to me 👍

PR #394, opened 2025-06-08
CI: Use hashes for GitHub actions

oscargus:
See https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash for motivation why this is the safer way to handle action versions.

Also update the brunch depedency version.

One may consider activating the dependabot action, see https://github.com/apytypes/apytypes/blob/main/.github/dependabot.yml for an example (although one may consider checking weekly rather than monthly). This will create a PR to update the versions/hashes to the latest version when a new one is available.

### DIFF
--- a/.github/workflows/build-test-all.yml
+++ b/.github/workflows/build-test-all.yml
@@ -43,7 +43,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


Schottkyc137:
This looks good to me in theory but the clear disadvantage of the approach is readability. Is there some real danger that one of the actions could be overtaken by a malicious actor?

PR #424, opened 2026-01-03
Move to use workspace dependencies and remove pad dependency

### LINKED ISSUE #401: Use of "Pad" dependency
I was just curious why there is a dependency on "Pad" when the Rust format command looks like it might be able to fulfill the need without pulling in another crate? Happy to push a PR to this effect, but thought that I would see if I am missing something obvious first.

https://github.com/VHDL-LS/rust_hdl/blob/4f20cbded129a2ed44e245de49907e5b1da69b8a/vhdl_lang/src/data/source.rs#L363

https://doc.rust-lang.org/std/fmt/#fillalignment

It seems like this could be done without an extra crate, basically. Especially for just the one usage.

oscargus:
Although there are not that many overlapping dependencies between the crates, it still gives a better overview if all are available in the workspace Cargo.toml when it comes to updating etc.

Removed the use of the `pad` crate. Closes #401 (not sure if it is tested, but should be the same if I understand the code).

Also updated a few dependencies, primarily to get rid of some transitive dependencies.


vhdl_lang/tests/integration_tests.rs
```diff
@@ -46,7 +46,7 @@ pub fn parses_example_project_without_errors() {
 
 #[test]
 fn unused_function_gets_detected() -> Result<(), Box<dyn Error>> {
-    let mut cmd = Command::cargo_bin("vhdl_lang")?;
+    let mut cmd = Command::new(cargo::cargo_bin!("vhdl_lang"));
```
Schottkyc137: What is the advantage of this change? Better error reporting?

Schottkyc137 (approved):
Thanks! These changes look good to me.
I think the code should be tested since it's used in `SrcPos::show` which is tested in the `source.rs` file.

At some point, we should probably also add a coverage stage to the pipeline to have a better understanding of what code is actually tested...

PR #425, opened 2026-01-03
Fix some pedantic clippy warnings to reduce code length

oscargus:
Focused on things that resulted in less code.

### DIFF
--- a/vhdl_lang/src/syntax/tokens/tokenizer.rs
+++ b/vhdl_lang/src/syntax/tokens/tokenizer.rs
@@ -1188,7 +1188,7 @@ fn parse_quoted(
     };
 
     match quoted_inner() {
-        Ok(_) => {}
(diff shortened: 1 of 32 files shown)


vhdl_lang/src/syntax/tokens/tokenizer.rs
```diff
@@ -1188,7 +1188,7 @@ fn parse_quoted(
     };
 
     match quoted_inner() {
-        Ok(_) => {}
+        Ok(()) => {}
```
Schottkyc137: Why is this better? (Not against this change, just curious)

oscargus:
Sorry for not getting back on this. These are all pedantic clippy warnings, so I think that the things you asked about are mainly related to consistency and someone has decided that the suggested form is the preferred one.

PR #426, opened 2026-01-18
Add casing categories and enable per-category override

### LINKED ISSUE #422: Less greedy preferred-case
Hello,

The new option `preferred-case` is really nice and welcome!
Specifically for the standard library where almost all types are in capital letters.
However, when instantiating components or entities from my own libraries, it now sets all generics to lowercase (since I configured `preferred-case` to `"lower"`).

Is it possible to set different levels for this feature? Like constraining this feature only to standard types maybe?

Thank you!

Schottkyc137:
This PR makes the case transformations for completions more usable.

Closes #422 

# Old
Previously, there was one `preferred_case` option:

```toml
preferred_case = "..." # lower, upper, pascal
```

# New
This PR enables more elaborate casing rules. The following are available categories:
- `constant` for constants (also includes generics)
- `design_unit` for all design units like entities, architectures, packages, e.t.c.
- `object` for all non-constant objects, i.e., signals, variables, files
- `subprograms` for functions and procedures
- `type` for types
- `enum_literal` for enumeration literals
- `label` for labels
- `attribute` for attribute names, i.e., `foo'high`
- `library` for library names, i.e, `ieee`, `work`
- `keyword` for VHDL keywords
- `alias` for aliases

A user configures the case setting by selecting a preset and optionally overrides:
```toml
[casing]
preset = "vhdl-common"

[casing.overrides]
type = "pascal" # Use the `vhdl-common` preset, but use pascal-case for types
```

There are two presets in this initial release:
1) `vhdl-common`: More or less opinionated casing styles that seem to be common:
```toml
constant = "upper"
attribute = "lower"
design_unit = "lower"
enum_literal = "lower"
keyword = "lower"
label = "lower"
library = "lower"
object = "lower"
subprogram = "lower"
type = "lower"
alias = "lower"
```

2) `keep`: Keep every category. This is intended as a starting point:
```toml
constant = "keep"
attribute = "keep"
design_unit = "keep"
# ...
```

### DIFF
--- a/vhdl_lang/src/config/case.rs
+++ b/vhdl_lang/src/config/case.rs
@@ -267,11 +267,11 @@ mod tests {
     }
 
     #[test]
-    fn enabled_by_default_only() {
+    fn default_preset_no_overrides() {
         let cfg = parse(
             r#"
             [casing]
-            preset = "default"
+            preset = "vhdl_common"
             "#,
         )
         .unwrap()
@@ -285,7 +285,7 @@ mod tests {
         let cfg = parse(
             r#"
             [casing]
-            default = "lower"
+            preset = "vhdl_common"
 
             [casing.overrides]
             constant = "upper"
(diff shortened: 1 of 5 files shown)


nselvara:
Does this also ignore standard libraries like IEEE?

PR #427, opened 2026-01-31
Add the `vhdl-syntax` crate

Schottkyc137:
The `vhdl-syntax` crate contains methods to parse and analyze VHDL files.
It is intended as public-API for users to write all sorts of tooling around VHDL files.
Examples include documentation generators, linters, formatter, diagram generators, e.t.c.


nselvara:
No worries mate, no one can write an entire full suite of software in one shot. Keep ut up ya!
Thank you again for your contribution! 🚀

P. S. Are you German? 🙈

PR #450, opened 2026-03-31
Add semantic tokens support (textDocument/semanticTokens/full and /range)

### LINKED ISSUE #314: [FEATURE] Semantic Tokens
It would be very nice to get [Semantic Tokens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens) from the Language Server.
This would enable colored tokens in the client ;)

axlegrinder1:
Expose VHDL entity classification to the editor via LSP semantic tokens, enabling context-aware coloring for signals, variables, constants, types, functions, ports, generics, and other VHDL constructs.

Uses standard LSP token types (variable, parameter, function, type, etc.) for out-of-the-box theme compatibility. Constants and generics are distinguished via the readonly modifier.

Results are cached per file and invalidated on change.

Closes VHDL-LS/rust_hdl#314

I built this functionality pretty quickly alongside Claude Opus4.6, so I would consider this a conceptual implementation which likely has room for improvement, but the results are very promising. I have added a set of tests within the vhdl_server.rs module and run this functionality in vscode + teroshdl across several large projects (200-500 source files) with good results.

With the feature, once the server is running, the code editor has a lot more visibility within the files, giving valuable context that was previously missing, such as being able to colour valid constants/enums in line with code, not just at declaration.

This is currently working cleanly as a separate module utilising the searcher infrastructure, but a tighter integration into the analysis pass could eliminate the redundant walk and guarantee consistency but would be a far more involved change within vhdl_lang.


vhdl_ls/src/vhdl_server.rs
```diff
@@ -1005,4 +1009,221 @@ lib.files = [
...
+
+    fn std_lib_config() -> String {
+        format!(
+            "[libraries]\nstd.files = ['{}/../vhdl_libraries/std/*.vhd']\nlib.files = ['*.vhd']\n",
+            std::env::var("CARGO_MANIFEST_DIR").unwrap()
+        )
+    }
+
+    /// Decode delta-encoded semantic tokens to (line, start, length, token_type, modifiers).
+    fn decode_semantic_tokens(tokens: &[SemanticToken]) -> Vec<(u32, u32, u32, u32, u32)> {
```
Schottkyc137: Please return a more meaningful data type. A tuple that consists of `n` `u32`s is really hard to read.

vhdl_ls/src/vhdl_server/text_document.rs
```diff
@@ -41,6 +43,7 @@ impl VHDLServer {
                 source.change(range.as_ref(), &content_change.text);
             }
             self.project.update_source(&source);
+            self.semantic_token_cache.clear();
```
Schottkyc137, line 48: Feels weird that the entire cache is cleared (which contains all URIs) if only a single file is updated. Is there any reason for this? (See also above)

vhdl_ls/src/vhdl_server/semantic_tokens.rs
```diff
@@ -0,0 +1,203 @@
...
+            .cmp(&b.0.start.line)
+            .then(a.0.start.character.cmp(&b.0.start.character))
+    });
+
+    tokens
+}
+
+/// Delta-encode sorted tokens, optionally filtering to a range.
+fn encode(
+    tokens: &[(vhdl_lang::Range, u32, u32)],
```
Schottkyc137: This tuple `(Range, u32, u32)` is not descriptive. I suggest changing it to some small struct.

vhdl_ls/src/vhdl_server/semantic_tokens.rs
```diff
@@ -0,0 +1,203 @@
...
+            Some((range, token_type, token_modifiers))
+        })
+        .collect();
+
+    tokens.sort_by(|a, b| {
+        a.0.start
+            .line
+            .cmp(&b.0.start.line)
+            .then(a.0.start.character.cmp(&b.0.start.character))
+    });
```
Schottkyc137: IIRC `SrcPos` already implements `Ord` so I think this is a bit overcomplicated

vhdl_ls/src/vhdl_server/semantic_tokens.rs
```diff
@@ -0,0 +1,203 @@
...
+        AnyEntKind::Concurrent(..) | AnyEntKind::Sequential(..) => return None,
+    };
+    Some(result)
+}
+
+/// Check if a token overlaps the filter range by line.
+/// Character-level precision is not needed as clients request full-line ranges.
+fn in_range(token_range: &vhdl_lang::Range, filter: &vhdl_lang::Range) -> bool {
+    token_range.start.line <= filter.end.line && token_range.end.line >= filter.start.line
+}
```
Schottkyc137: If this kind of utility doesn't exist in `vhdl_lang::Range` already I suggest to move it there

vhdl_ls/src/vhdl_server/semantic_tokens.rs
```diff
@@ -0,0 +1,203 @@
...
+        Type::Enum(_) => (ENUM, 0),
+        Type::Record(_) => (STRUCT, 0),
+        Type::Protected(..) => (CLASS, 0),
+        Type::Subtype(sub) => type_token(sub.type_mark().kind()),
+        Type::Alias(t) => type_token(t.kind()),
+        _ => (TYPE, 0),
+    }
+}
+
+fn to_semantic_token(kind: &AnyEntKind) -> Option<(u32, u32)> {
```
Schottkyc137: The `(u32, u32)` is again not really descriptive. It's OKish here compared to the examples down, but I prefer if this is some small descriptive `struct`.

vhdl_ls/src/vhdl_server/semantic_tokens.rs
```diff
@@ -0,0 +1,203 @@
...
+    SemanticTokenType::TYPE,        // 5: types (general)
+    SemanticTokenType::CLASS,       // 6: protected types, components
+    SemanticTokenType::NAMESPACE,   // 7: libraries, design units, labels
+    SemanticTokenType::STRUCT,      // 8: record types
+    SemanticTokenType::ENUM,        // 9: enum types
+];
+
+pub const TOKEN_MODIFIERS: &[SemanticTokenModifier] = &[
+    SemanticTokenModifier::READONLY, // bit 0: constants, generics
+];
```
Schottkyc137: I think this should be a little macro (or similar) to make extending the semantic token types future-proof

vhdl_lang/src/project.rs
```diff
@@ -335,6 +335,14 @@ impl Project {
         self.root.find_all_references_in_source(source, ent)
     }
 
+    /// Collect all (position, entity) pairs in a source file for semantic token support.
+    pub fn semantic_tokens(&self, source: &Source) -> Vec<(SrcPos, EntRef<'_>)> {
```
Schottkyc137: This shouldn't be called `semantic_tokens`. After all it just collects all tokens and associates them with the entity they belong to.
The language server than simply decides to use that information to compute semantic tokens.
I suggest giving this a more appropriate title.

vhdl_lang/src/ast/search.rs
```diff
@@ -2005,6 +2006,52 @@ impl Searcher for FindAllUnresolved {
     }
 }
 
+/// Collects all (position, entity) pairs in a source file for semantic token support.
+pub struct SemanticTokenCollector<'a> {
```
Schottkyc137, line 2010: It somehow strikes me as weird that all this is needed just to combine each token (`SrcPos`) with an `EntRef` - this is a layer that `vhdl_lang` should already have (it's likely just not exposed / hidden).
If there is a better solution I'm open for improvements that are more intrusive.
But that being said, maybe I'm wrong and this truly is the only sensible way to combine this information.

Schottkyc137 (changes requested):
Thanks for tackling this!
Implementation looks overall like a good starting point. I have left a few comments.
Concerning more involved changes of `vhdl_lang`: I have nothing against that if there is real advantage. Just please make sure to structure this, e.g., by opening multiple PRs such that it's easier to review.

vhdl_lang/src/ast/search.rs
axlegrinder1: I looked for an existing way to get all (position, entity) pairs from a source file. The resolved references are stored as Reference fields on individual AST nodes (WithRef<T>, WithDecl<T>) — there's no pre-built collection of them. The Searcher visitor is the existing mechanism to access them in bulk (all 12 existing implementations — FindAllReferences, FindAllUnresolved, CompletionSearcher, etc. — use the same pattern). SemanticTokenCollector is essentially the minimal Searcher that collects everything without filtering. If there's a better entry point I've missed I'd love to refactor with this.

vhdl_ls/src/vhdl_server/text_document.rs
axlegrinder1: It's due to cross-file references. I added a clarifying comment. It's a bit heavy handed maybe but ensures no stale tokens exist for any file.

vhdl_ls/src/vhdl_server/semantic_tokens.rs
axlegrinder1: Yep, missed that. Refactored to use Ord

vhdl_ls/src/vhdl_server/semantic_tokens.rs
axlegrinder1: I've moved the function to the Range::overlaps_lines

vhdl_ls/src/vhdl_server/semantic_tokens.rs
axlegrinder1: I added a define_token_types! macro that generates both the index constants and the TOKEN_TYPES legend array from a single declaration, so now they can't get out of sync when extending

vhdl_lang/src/project.rs
axlegrinder1: Good point, renamed to `find_all_entity_references` to match naming convention of similar functions

vhdl_lang/src/ast/search.rs
Schottkyc137, line 2010: This should now also be renamed (similar to `semantic_tokens` -> `find_all_entity_references`).

axlegrinder1:
@Schottkyc137 Your comments have been addressed, is there any more that you want to see in this feature or is it just in the backlog now? I'm eager to go further with development of this if needed, but else I'm very happy with the result and would appreciate feedback :) I have been using this build day-to-day and it has been very stable in my own use-case.

Schottkyc137 (approved):

Schottkyc137:
Hi
Sorry for the long delay. All looks good with this implementation, I'll merge this in a bit.

PR #452, opened 2026-04-06
Add a checked API

Schottkyc137:
`CheckedNode`s are a wrapper around `SyntaxNode`s with guarantees that there are no missing or extraneous elements.
This information is extracted from the new meta-API and checked nodes can only be constructed (safely) if they pass validation.
They enable a more ergonomic API entry to the design as all optional nodes are now truly optional, and all required nodes simply return the node instead of `Option` as would be the case for the `SyntaxNode`

Changes are fairly simple, but contains ~10k of generated code

### DIFF
--- a/vhdl_syntax/src/builder.rs
+++ b/vhdl_syntax/src/builder.rs
@@ -102,6 +102,18 @@ impl From<&[u8]> for Identifier {
     }
 }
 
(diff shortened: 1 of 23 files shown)


Schottkyc137:
Superseeded by #509

PR #459, opened 2026-04-26
fix: switch to gen-lsp-types

ribru17:
[gen-lsp-types](https://github.com/ribru17/gen-lsp-types) is an alternative to the lsp-types crate, with types generated via codegen from the official LSP Metamodel for correctness and completeness.

lsp-types issues fixed in gen-lsp-types:

- https://github.com/gluon-lang/lsp-types/issues/310
- https://github.com/gluon-lang/lsp-types/issues/308
- https://github.com/gluon-lang/lsp-types/issues/284
- https://github.com/gluon-lang/lsp-types/issues/278
- https://github.com/gluon-lang/lsp-types/issues/277
- https://github.com/gluon-lang/lsp-types/issues/260
- https://github.com/gluon-lang/lsp-types/issues/245
- https://github.com/gluon-lang/lsp-types/issues/93

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,12 +108,6 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bitflags"
-version = "1.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
-
-[[package]]
-name = "bitflags"
 version = "2.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


Schottkyc137:
Hi and thanks for the PR!

I have checked out your project and I really like it. Your design decisions look solid and some of the pain points you address are ones I've also hit in `lsp_types` (the `uri` vs `url` decision being the main one).

That being said, my major concern with your PR is stability: I don't have an immediate problem with `lsp_types` and it's battle-tested. Your new crate has 5 versions in roughly two weeks. May I ask what your plan for stability is?

In any case, I'm strongly considering switching — either when I hit a feature `lsp_types` doesn't support, or once your crate has matured a bit.

ribru17:
Thanks for the feedback! Lack of stability is a very valid point. In that case, maybe you should wait on this PR. I *think* there shouldn't be any more breaking versions coming in the future, aside from those due to actual LSP metamodel updates. That said, there might be something that comes up which requires a breaking fix, and just in case maybe we should hold off on this? That said, I'll try my best not to have this be the case, especially since the library is used in a couple of projects now.

PR #475, opened 2026-05-28
Add goto_type_definition and reload the project if a file was created/renamed/deleted

kernmatthias:
When using vhdl_ls I came across a few issues that should be fixed with this PR:
- goto_type_definition support: allow to go from a name/selected name to go directly to the type of the signal or variable
- handling Did{Rename,Create,Delete}File so that these events cause a reload of the project, so adding or renaming of files does not require a restart of the language server to have the diagnostics working in the new file


Schottkyc137:
Many thanks! This looks great!

PR #507, opened 2026-09-13
VHDL-Lint: A modern VHDL linter

Schottkyc137:
VHDL-Lint is a modern linter for VHDL that takes inspiration from popular linters like ruff, clippy and ESLint. This is the initial implementation with a minimal feature set.


teburd:
awesome

